PR #5, opened 2015-06-04
Added Integration tests

agenthunt:
Hi,

I added some integration tests for the existing api. Let me know your comments.

Thanks,
Shailesh.


johanneslumpe:
Thanks for the PR - checked out the tests, nice! I noticed that if I let a test fail on purpose, the test-runner never finishes. Have you encountered that behavior too?

agenthunt:
I have noticed that if done() function is not called at the end of a test, it doesn't finish.

johanneslumpe:
Ok, can you fix that then? :)

agenthunt:
Sure, which test case was that you made it fail ?

johanneslumpe:
I just changed the [value to compare to](https://github.com/johanneslumpe/react-native-fs/pull/5/files#diff-d5efcfd23472e6b30047120837211886R55) to a static value, in order to see what would happen if it failed. The test itself doesn't fail normally, but it would be nice to get this fixed anyway instead of having hanging tests on errors.

agenthunt:
Thanks. Fixed that now. Realised that there was another error that " test message was not showing in Xcode". Fixed that as well.

johanneslumpe:
Ok cool! Let me merge this in for now :)

agenthunt:
Thanks!!!

johanneslumpe:
No thank you :)

PR #7, opened 2015-06-10
Get rid of peerDependencies

jtremback:
I see no need for this and it is preventing me from using this module. I need to make this fork anyway, so i thought i'd make a PR as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,5 @@
     "base-64": "^0.1.0",
     "bluebird": "^2.9.25"
   },
-  "peerDependencies": {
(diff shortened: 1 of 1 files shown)


johanneslumpe:
That's fine with me for now, but maybe we could mark it as a dev dependency so that you can still do `npm i` when you want to build the static library on its or when you want to run the integration tests. Could you add it with a  `*` there?

johanneslumpe:
ping @jtremback Will you change this in the PR or do you want me to do this afterwards?

johanneslumpe:
Ah well, nevermind.

PR #8, opened 2015-06-12
First stab at node like api

agenthunt:
Hi,

I took a first stab to make the existing api node-like.   I hope it can serve as a good starting point for discussion/further changes. Please have a look and let me know your feedback.

Thanks,
Shailesh.

### DIFF
--- a/FS.ios.js
+++ b/FS.ios.js
@@ -2,13 +2,7 @@
 
 var RNFSManager = require('NativeModules').RNFSManager;
 var Promise = require('bluebird');
-var base64 = require('base-64');
 
-var _readDir = Promise.promisify(RNFSManager.readDir);
-var _stat = Promise.promisify(RNFSManager.stat);
-var _readFile = Promise.promisify(RNFSManager.readFile);
-var _writeFile = Promise.promisify(RNFSManager.writeFile);
-var _unlink = Promise.promisify(RNFSManager.unlink);
 
 var convertError = (err) => {
   if (err.isOperational) {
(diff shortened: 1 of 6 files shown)


johanneslumpe:
I will take a proper look at this soon. In the meantime, can you fix the indentation? 2 spaces and please don't indent the returned dictionaries. (I know Xcode is like that). Please try to keep the original code style. Thanks :)

agenthunt:
Absolutely. Sorry about the formatting messup. Mind if I ask you what is your dev setup ?. I tried to use sublime + JS prettify, but it doesnt seem play nicely with es6 syntax and jsx. At present using atom.

johanneslumpe:
Using atom + Xcode here.

agenthunt:
ok. great. I am on the same setup . Was there any formatting issues with javascript or was it only objective c code  ?

johanneslumpe:
There are indentation issues here:
https://github.com/agenthunt/react-native-fs/blob/topic/fsapi/FS.ios.js#L61
https://github.com/agenthunt/react-native-fs/blob/topic/fsapi/FS.ios.js#L126

Excessive new lines:
https://github.com/agenthunt/react-native-fs/blob/topic/fsapi/FS.ios.js#L5
https://github.com/agenthunt/react-native-fs/blob/topic/fsapi/FS.ios.js#L156

Here there are two unnecesary `else` blocks:
https://github.com/agenthunt/react-native-fs/blob/topic/fsapi/FS.ios.js#L42
https://github.com/agenthunt/react-native-fs/blob/topic/fsapi/FS.ios.js#L46

I think there are places which can be made cleaner (https://github.com/agenthunt/react-native-fs/blob/topic/fsapi/FS.ios.js#L52 for example. Instead of having the call duplicated, just create an args-array and `apply` it at the end) but that's stuff we can discuss later :)

I'm not sure whether I want the promisified api to live in the `Promisified` namespace. That turns me off a bit - bit I haven't thought of a better solution yet:D

agenthunt:
Thanks for the detailed feedback. Are you using atom-beautify ? If yes, Can you share your .jsbeautifyrc file? May be we can include that in the root of the project ?

johanneslumpe:
Not using it currently, but I could set it up I guess :)

johanneslumpe:
@agenthunt I think your last commit made the identation just worse ;)

agenthunt:
What? seriously ? I changed xcode indent to 2 spaces. Using atom with atom-beautify. Seems like this doesnt like es6 syntax as well.   Please let me know your settings. I will try to use that.

johanneslumpe:
@agenthunt I still haven't set up atom-beautifier, but in general: indentation of 2 spaces that' the biggest issue here ;)

johanneslumpe:
@agenthunt that looks much better - somehow I didn't get a notification from github, that's why I'm replying so late :/

johanneslumpe:
This will be a breaking change, so let's make it a good as possible before we merge this in!

FS.ios.js
```diff
@@ -23,48 +16,138 @@ var convertError = (err) => {
...
+ * Ex: Documents/hello.txt -> <DocumentsDirForApp>/hello.txt
+ * Caches/hello.txt -> <CacheDirForApp>/hello.txt
+ * tmp/hello.txt -> <TmpDirForApp>/hello.txt
+ * @param  {String}
+ * @return {String}
+ */
+function processFilePath(filepath) {
+  var found = filepath.indexOf('/');
+  if (found !== 0) {
+    //relative path, check where is it relative to
```
johanneslumpe: Using a `switch` statement here would clean thing up a bit:

```javascript
var splitPath = filepath.split('/');
switch (splitPath[0]) {
  case 'Documents':
    splitPath[0] = RNFSManager.NSDocumentDirectoryPath;
    break;
  case 'Caches':
    splitPath[0] = RNFSManager.NSCachesDirectoryPath;
    break;
  case 'tmp':
    splitPath[0] = RNFSManager.NSTmpDirectoryPath;
    break;
  default:
    return RNFSManager.NSDocumentDirectoryPath + '/' + filepath;
}

return splitPath.join('/');
```

Seems much more readable to me.

FS.ios.js
```diff
@@ -23,48 +16,138 @@ var convertError = (err) => {
...
 
-  stat(filepath) {
-    return _stat(filepath)
-      .then((result) => {
-        return {
-          'ctime': new Date(result.ctime*1000),
-          'mtime': new Date(result.mtime*1000),
-          'size': result.size,
-          'mode': result.mode,
+  readFile(filepath) {
```
johanneslumpe: I don't like that we are not specifying all arguments here. I'd rather have something like:

`readFile(filepath, options, callback)`

and then a check like this: 

```
readFile(filepath, options, callback) {
  var processedFilePath = processFilePath(filepath);
  var hasCallback = typeof callback === 'function';
  callback = hasCallback ? callback : options;
  options = hasCallback ? options : DEFAULT_OPTIONS;
  
  RNFSManager.readFile(processedFilePath, options, callback);
}
```

FS.ios.js
```diff
@@ -23,48 +16,138 @@ var convertError = (err) => {
...
+      }, callback);
+    }
+  },
+
+  readdir(filepath, callback) {
+    var processedFilePath = processFilePath(filepath);
+    RNFSManager.readdir(processedFilePath, callback);
+  },
+
+  mkdir(filepath) {
```
johanneslumpe: Same as for `readFile`:

`mkdir(filepath, mode, callback)`

And then also an updated check for the callback.

agenthunt:
@johanneslumpe Thanks. Agreed. should be good as possible.  Am a little occupied till Monday. Will take a look at it then.

agenthunt:
Not sure what happened with xcode setup. bumping into this error. Tried a bunch of stuff from google. Unable to fix. Any thoughts ?

Undefined symbols for architecture i386:
  "_RCTRegisterModule", referenced from:
      +[RNFSManager load] in libRNFS.a(RNFSManager.o)
ld: symbol(s) not found for architecture i386
clang: error: linker command failed with exit code 1 (use -v to see invocation)

johanneslumpe:
@agenthunt hmm that seems rather odd. Maybe this helps: http://stackoverflow.com/questions/6984368/undefined-symbols-for-architecture-i386-objc-class-skpsmtpmessage-referenc

johanneslumpe:
Closing as inactive

PR #10, opened 2015-08-17
make compatible with react-native 0.10.0

zh217:
see the breaking change concerning NSNumber at https://github.com/facebook/react-native/releases/tag/v0.10.0-rc


zh217:
I see, this is a duplicate

PR #14, opened 2015-08-27
Update to load Bundles by their class name if not found in typical bundle path.

MattFoley:
This is mainly for dynamic/embedded frameworks. While they can be loaded as an NSBundle, they're not in the typical path.


jnhuynh:
Looks good to me.

MattFoley:
@johanneslumpe If you've got any feedback on this, I'll be happy to implement it, but I'm going to go ahead and merge this now since we're on radically different time zones.

johanneslumpe:
Just published with version 0.4.0!

MattFoley:
Cool! Will update to that. Been swamped myself.

PR #15, opened 2015-10-01
Requiring internal react-native modules will stop working

jamesfzhang:
`require('NativeModules')` now creates this warning:

```
Unable to resolve module NativeModules from /Users/jameszhang/Documents/thedrop/TheArtistUnion/node_modules/react-native-fs/FS.ios.js
```

And should be replaced with `require('react-native').NativeModules`

Issue originally reported here: https://github.com/facebook/react-native/issues/1808

### DIFF
--- a/FS.ios.js
+++ b/FS.ios.js
@@ -1,6 +1,6 @@
 'use strict';
 
-var RNFSManager = require('NativeModules').RNFSManager;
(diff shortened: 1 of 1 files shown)


johanneslumpe:
Thanks!

johanneslumpe:
Just published with version 0.4.0!

PR #28, opened 2015-11-23
Feature/stop file download

MiincGu:
Add ability to stop file download.


cjdell:
@MiincGu Thanks for doing this. I'll give this some testing later today and if all goes well I will merge this in. :-)

cjdell:
@MiincGu Again thanks for you contribution. There is a slight issue in that your implementation assumes that only one download can be happening at any one time. If two download jobs are running concurrently then your `stopDownload` will only abort the last one. It has however made me realise that the exposed API could be better.

I'm going to make another callback that will return the job ID of the download as soon as the download begins, along with other useful infos like response headers. The caller will be able to use the job ID to tell the module which download to abort.

cjdell:
@MiincGu Have refactored and manually merged in some of your code. Need to test `stopDownload` in new implementation but I've run out of time today. This pull request is now redundant.

MiincGu:
Thanks for your job.

PR #29, opened 2015-11-24
Feature/download error handler

MiincGu:
Merged latest code in master branch, and did a little change.

This change make it possible to handle error outside the plugin, useful in some case.


cjdell:
@MiincGu I'm not sure why this is necessary to be honest. Errors are passed from native by calling the callback with an object as the first argument (Node.js style error callback). Because the calls to native are promisified, this means the promise will reject if the native function calls back with an error.

MiincGu:
Sorry for my misunderstanding.

Just close it.

PR #40, opened 2016-01-07
Add CocoaPods support

oblador:
This PR adds a podspec file so that the library can be integrated using cocoapods.


johanneslumpe:
Cool, thanks!

PR #43, opened 2016-01-16
add move file method

asamiller:
This adds a new method for moving a file (or renaming a file). It's only iOS though (I don't really know java). 

```javascript
 RNFS.moveFile(sourcePath, destPath);
```


johanneslumpe:
@cjdell Do you think you can provide the Android equivalent for this?

Issue #44, opened 2016-01-20
downloadFile fail will cause app crash in iOS


magicismight:
When download file from url failed,will cause app crash in iOS.

```
return callback(@[[NSNull null], @{@"jobId": jobId,
```

will throw an erro 
Threa 1: signal SIGABRT

PR #45, opened 2016-01-20
fix error caused by when `bytesWritten` is null

magicismight:
Fix error  caused by when  `bytesWritten` is null.
issue #44


magicismight:
It will occur when tIt will occur when the HTTP response failed.he HTTP response failed.

cjdell:
Thanks for finding this :-)

PR #48, opened 2016-01-28
add exist method

iwater:


corbt:
This is a very useful function, but I think it would be better/more standard if it were named `exists` instead of `exist`.

johanneslumpe:
@cjdell Could you give this a test run? I'm currently occupied with other work ;(

corbt:
@johanneslumpe I'm using this fork on iOS right now and I can verify that it works well on that platform. Haven't tried on Android. I still think like I mentioned earlier that the method would be better named `exists` than `exist` for consistency with other similar libraries.

iwater:
@corbt it's named exists	 now :)

corbt:
looks good to me :+1:

cjdell:
@johanneslumpe I'll update the demo app and do some tests when I get a moment, been meaning to do this for a while. I'm a little busy myself :-)

johanneslumpe:
@corbt @iwater sorry for the long wait. Have you tested this for iOS and Android? If so I will merge this in :) just give me a short feedback please.

corbt:
I've tested on iOS and it works.

iwater:
I've tested on iOS and Android before created this PR.

nicolasiensen:
I'm getting the following error

```undefined is not a function (evaluating '_reactNativeFs2.default.exist...```

The line where this exception is being raised looks like this:

```RNFS.exist(`/mnt/sdcard/${track.id}.${track.original_format}`)```

In my `package.js` I've pointed the library to the Github repo:

 `"react-native-fs": "johanneslumpe/react-native-fs"`

Thoughts?

PR #49, opened 2016-02-06
Support file:// and library content:// files

jrichardlai:
Not sure if that's the best way, but it allows reading file from library or filesystem.

### DIFF
--- a/android/src/main/java/com/rnfs/RNFSManager.java
+++ b/android/src/main/java/com/rnfs/RNFSManager.java
@@ -5,6 +5,7 @@ import java.util.Map;
 import java.util.HashMap;
 import java.util.ArrayList;
 
+import android.net.Uri;
 import android.os.Environment;
 import android.os.AsyncTask;
 import android.util.Base64;
@@ -14,10 +15,8 @@ import android.util.SparseArray;
 
 import java.io.File;
 import java.io.FileOutputStream;
-import java.io.FileInputStream;
-import java.io.BufferedInputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.InputStream;
-import java.io.OutputStream;
 import java.io.IOException;
 import java.net.URL;
 import java.net.URLConnection;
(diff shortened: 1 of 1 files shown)


johanneslumpe:
@jrichardlai Thanks! I'm not an Android-person but I guess @cjdell has an opinion on this :)

johanneslumpe:
@jrichardlai Are you using this in production? I'm inclined to merge this as @cjdell isn't responding. But I cannot test it on my own right now, so I want to make sure that you are positive that this works as expected. Is there anything you think we should add to the readme/docs for this?

jrichardlai:
@johanneslumpe I'm using it in production didn't see any problems so far, but this feature is not used a lot. Also I added this change because it makes sense for us ( we use it when we want to retry an upload that fails, which can be from either a image from the library or a new photo ) but I don't know if it fits in this library if react-native-file-system is really about only the file system.

jrichardlai:
And yeah a second opinion would be better :)

mcorb:
Accepting a subset or URLs using existing entry points is confusing and could introduce correctness issues because this API is specified as accepting file paths.

(e.g. why don't the other functions in the API accept URLs, only this one? Why are only some kinds of URLs accepted? What if I really want to write a file that looks like a URL in some custom Android distribution, but this code prevents me from doing so?)

jrichardlai:
@mcorb word. Maybe I should make another method for that `readFileFromURL`? Since this library is already handling download of urls, it still makes sense to be part of it.

bnns:
@grokbot Thanks for posting your finding about `ExternalDirectoryPath` - it just saved me a lot of time!

Froelund:
This works for me. Thanks @jrichardlai 
In my scenario I tried to read the output file from [react-native-camera](https://github.com/lwansbrough/react-native-camera). It worked on ios, but failed on Android. This fixed the issue 👍

matto1990:
This certainly fixes and issue with Android, but I agree that it should be implement on all methods so the support is consistent.

jjzazuet:
Hi. Just pinging to check if any decision has been made on this PR. Thanks!

android/src/main/java/com/rnfs/RNFSManager.java
```diff
@@ -75,15 +123,13 @@ public void writeFile(String filepath, String base64Content, ReadableMap options
   @ReactMethod
   public void readFile(String filepath, Callback callback) {
     try {
-      File file = new File(filepath);
-
-      if (!file.exists()) throw new Exception("File does not exist");
```
itinance: Can you add please again the File-Exists-check and throw an Exception, if the file does not exists? To not throw an exception would be compatibility break for all those that rely on this test (instead of calling .exists() by themself in JS-layer)

itinance:
Thanks for reaching out. Just one thing. Please consider my comment

android/src/main/java/com/rnfs/RNFSManager.java
ItsNoHax: I would also recommend not adding file:// without checking if the filepath already starts with file:// to avoid double file://file://

wassgha: It actually doesn't start with `file://` since the uri scheme is null

TomMahle:
@grokbot Could you elaborate on this? I'm using:
`Linking.openURL('file://' + RNFS.ExternalDirectoryPath + '/' + fileName)`
but the file is still not being opened even though Linking.canOpenURL is giving back true and PermissionsAndroid.PERMISSIONS.READ_EXTERNAL_STORAGE is giving granted. Any ideas what I might still be missing?

PR #51, opened 2016-02-07
Update Readme.md re: new react-native version

scottmmjackson:
react-native now uses `getPackages()` to return an array of package objects instead of overriding `onCreate()`.


johanneslumpe:
//cc @cjdell

johanneslumpe:
Thanks!

PR #61, opened 2016-03-18
new getFSInfo command

corbt:
This creates a new `getFSInfo()` command. Its call signature is as follows (which I've also added to the docs):

### `promise getFSInfo()`

Returns an object with the following properties:

`totalSpace` (`Number`): The total amount of storage space on the device (in bytes).
`freeSpace` (`Number`): The amount of available storage space on the device (in bytes).

I've tested this in iOS and it works fine. Haven't had a chance to test in Android because I'm having emulator issues, but it is implemented. If someone else could test that I'd appreciate it.


johanneslumpe:
@corbt Thanks! As soon as somebody is willing/able to test this on Android, I will merge it in.

johanneslumpe:
@cjdell Can you maybe give this a try? Would like to get this merged in too :)

cjdell:
@corbt Manually merging this. Thanks for your contribution :-)

PR #62, opened 2016-03-19
downloadFile: avoid connecting twice for each download on Android

mcorb:
A connection already exists to fetch HTTP headers, so use that to complete the
download instead of making a second connection.


johanneslumpe:
Thanks!

PR #66, opened 2016-03-31
Give android access to getExternalFilesDir

outofculture:
Much like e.g. `RNFS.DocumentDirectoryPath`, use [getExternalFilesDir](http://developer.android.com/reference/android/content/Context.html#getExternalFilesDir(java.lang.String)) to get the abs path to shared storage. Set to null for iOS or when there is no external storage.

Motivation:
> External storage is the best place for files that don't require access restrictions and for files that you want to share with other apps or allow the user to access with a computer.

### DIFF
--- a/IntegrationTests/FSTest.js
+++ b/IntegrationTests/FSTest.js
@@ -63,7 +63,7 @@ function testWriteAndReadFile() {
 
 
 function testCreateAndDeleteFile() {
-  var path = RNFS.ExternalDirectoryPath + '/test.txt';
+  var path = RNFS.DocumentDirectoryPath + '/test.txt';
   var text = 'Lorem ipsum dolor sit amet';
   var readText;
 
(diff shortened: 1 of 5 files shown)


johanneslumpe:
@cjdell This is another one to maybe give a test :)

liuchungui:
@johanneslumpe I need NSExternalDirectoryPath，can you merge it or add NSExternalDirectoryPath ?

IntegrationTests/FSTest.js
```diff
@@ -63,7 +63,7 @@ function testWriteAndReadFile() {
 
 
 function testCreateAndDeleteFile() {
-  var path = RNFS.DocumentDirectoryPath + '/test.txt';
+  var path = RNFS.ExternalDirectoryPath + '/test.txt';
```
johanneslumpe: This test does not work for iOS does it? Because `ExternalDirectoryPath` is `null`.

outofculture: Sorry, yeah. I'll fix that.

liuchungui:
@cjdell Thanks

PR #68, opened 2016-04-07
Add property jobId to progressCallback

### LINKED ISSUE #67: jobId in progressCallback of downloadFile
I'm actually working on a loader that start multiple downloads and save state into an object that contain all jobs, in purpose to do a loading screen.

I just noticed that `progressCallback` doesn't contain the property `jobId` as `beginCallback` does. It's fine when a single file is downloaded, but with multiple downloads we can't easily associate the progress callback with his job.

Is there a reason for not returning this property? It seems simple enough since `progressCallback` listen progression with the `jobId` on both platforms.

charpeni:
Fixed #67.

cc @cjdell.


johanneslumpe:
This looks good to me, let's see if @cjdell will comment here too :) Else I will merge this in soon.

Issue #75, opened 2016-04-14
confusing error message for readDir failure


outofculture:
Notably, `readDir('/storage/emulated/0/')` works just fine.

Reproduce with: https://github.com/outofculture/fsError

`adb logcat | grep -i boom` should show the error: `Error: Attempt to get length of null array`. You could get rid of the try/catch to get a stack trace, though it's only of the js, so it doesn't really show what's happening in the java.

(Sorry the example isn't exactly minimal; I'm also trying to track down a bug with what I think is redux interacting poorly with -fs.)

outofculture:
Oh! I don't know that the problem is the `0`-named directory; that's just a guess.

outofculture:
Ah ha! Directory permissions.

```
$ adb shell ls -l /storage/
drwx--x--x root     sdcard_rw          2016-01-29 11:36 emulated
```

No read permissions on the directory itself (even though some of its contents do have read permission, so it would allow reading of them). So this bug is really "Error message is confusing", and can be de-prioritized accordingly.

PR #77, opened 2016-04-15
Fixes error `attempt to insert nil object from objects[2] was thrown while invoking readDir on target`

jeduan:
This error happens when trying to read a file or directory for which the user doesn't have permissions. Doing this will throw the error below. This pull request skips over any file that throws an error when trying to read the attributes. Also skips over any dotfile found on the directory.

simulator_screen_shot_apr_13 _2016 _9 35 07_am

### DIFF
--- a/FS.common.js
+++ b/FS.common.js
@@ -158,9 +158,9 @@ var RNFS = {
   downloadFile(fromUrl, toFile, begin, progress) {
     var jobId = getJobId();
     var subscriptions = [];
-    var beginListener = begin || (info) => {
+    var beginListener = begin || ((info) => {
       console.log('Download begun:', info);
-    };
+    });
 
     if (begin) {
       subscriptions.push(NativeAppEventEmitter.addListener('DownloadBegin-' + jobId, begin));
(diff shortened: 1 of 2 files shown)


jeduan:
This is related to #75

johanneslumpe:
@jeduan I'm not sure that just skipping the files is expected behavior? This is like silently failing, which might be even more confusing to users?

jeduan:
Good point, should I rather break the loop and send an error the moment I get this?

johanneslumpe:
When this error occurs, that means that we cannot read any file in the directory, right? Because it would always break for those files. I think instead of skipping over those files or just failing hard, we might want to send an object with the filename and a custom error prop, so we can continue with the rest but the user will still know that the file failed. Then the user itself could filter the result to get rid of those files from the list maybe? What do you think about that? Just a suggestion though, nothing set in stone.

jeduan:
So the case I have is: (on iOS Simulator)

 - Read `/`
 - Will get info for every file/directory
 - When it reaches `/dev` it throws

That's why I just skipped over the directory that fails. Your solution seems reasonable, too. I can implement it if you want to go that way.

johanneslumpe:
@jeduan Did you ever get around to add the custom error object instead of just skipping over a file?

felgaaen:
Can you please merge this if the problem is solved? Still experiencing the issue

thanmaic:
Still experiencing the issue. Any fix or work around to this issue?
@jeduan @itinance 

RN version: 0.63.3
react-native-fs version: 2.13.3

jeduan:
This is over 4 years old. @thanmaic feel free to open a new PR based on this.

PR #85, opened 2016-04-29
Better Cocoapods integration

dabrut:
By declaring React/Core as a dependency we can easily add react-native-fs to our podfile


johanneslumpe:
@dabrut Is there any difference in declaring `React/Core` vs `React` as in  https://github.com/johanneslumpe/react-native-fs/pull/82 ?

dabrut:
It's pretty much the same except `React` is deprecated 
screenshot

PR #86, opened 2016-05-06
Add support for following 301/302 redirects on android

prha:
This is to address the issue outlined in https://github.com/johanneslumpe/react-native-fs/issues/71, where redirects are not followed in android.

The solution is similar to the one described in the comment section of the issue, but does not recursively call itself on the redirect url (avoids the danger of a stack overflow), and handles both 301/302 status codes.


johanneslumpe:
@cjdell got time to test this?

PR #88, opened 2016-05-14
Update RNFS.podspec

grabbou:
Support for CocoaPods 1.

Also, a side question - are there plans to push tags as well as bump versions in package.json? Currently it says 1.3.0 whereas the latest release on Github is 1.4.0.

### DIFF
--- a/RNFS.podspec
+++ b/RNFS.podspec
@@ -1,11 +1,19 @@
+require 'json'
+pjson = JSON.parse(File.read('package.json'))
+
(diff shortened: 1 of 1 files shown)


9mm:
+1 definitely need this

johanneslumpe:
@grabbou I think I just screwed up the github release ;) I usually bump versions in the package.json. Somehow something went wrong with that release, not sure what. But yes, those should be in sync...

PR #89, opened 2016-05-14
Promisify interface

grabbou:
Fixes #87 

Removes bluebird dependency, cleans up errors and some indent issues (aligned to 2 spaces).

Errors despite the change should work just fine, here's a sample error from iOS when reading directory that does not exist:
```
Error: The folder “dasdasdas” doesn’t exist.
    at createErrorFromErrorData (http://localhost:8081/index.ios.bundle?platform=ios&dev=true:3469:11)
    at http://localhost:8081/index.ios.bundle?platform=ios&dev=true:3424:11
    at MessageQueue.__invokeCallback (http://localhost:8081/index.ios.bundle?platform=ios&dev=true:3301:10)
    at http://localhost:8081/index.ios.bundle?platform=ios&dev=true:3184:8
    at guard (http://localhost:8081/index.ios.bundle?platform=ios&dev=true:3129:1)
    at MessageQueue.invokeCallbackAndReturnFlushedQueue (http://localhost:8081/index.ios.bundle?platform=ios&dev=true:3183:1)
    at DedicatedWorkerGlobalScope.onmessage (http://localhost:8081/debuggerWorker.js:39:56)
```
```js
console.log(err.code);
// ENSCOCOAERRORDOMAIN260

console.log(err.message);
// The folder “dasdasdas” doesn’t exist
```

The `err.code` is a pattern used in React Native for creating errors.

PS. Append `?w=false` to diff to ignore indent/whitespace changes.

PR #91, opened 2016-05-14
Fixes #90

### LINKED ISSUE #90: Subscriptions are likely never removed
https://github.com/johanneslumpe/react-native-fs/blob/master/FS.common.js#L170

If both being and progress are registered, the latter will overwrite the former which causes the subscriptionAndroid.remove() to only clear the last assigned.

We should use an array instead. We should also stick to one emitter regardless of the platform (NativeAppEventEmitter), I am going to send a patch for that as well.

grabbou:
Merge after #89 - needs rebasing after that due to indent changes.

### DIFF
--- a/FS.common.js
+++ b/FS.common.js
@@ -8,8 +8,7 @@ if (typeof self === 'undefined') {
 }
 
 var RNFSManager = require('react-native').NativeModules.RNFSManager;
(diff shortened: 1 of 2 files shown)


grabbou:
CC: @johanneslumpe you can merge that one first and I'll rebase #89 as it;s likely to take some time to test it out.

PR #93, opened 2016-05-14
Fixes #76

### LINKED ISSUE #76: Access to /tmp directory?
Hi, 

Thanks for the great work! It's been very helpful so far. 

I was wondering if there is a way to access the /tmp directory?

Thanks,
Andrea

grabbou:
AFAIK there's no tmp folder on Android (and it's advised to just use cache dir that we already expose).

### DIFF
--- a/android/src/main/java/com/rnfs/RNFSManager.java
+++ b/android/src/main/java/com/rnfs/RNFSManager.java
@@ -41,6 +41,7 @@ public class RNFSManager extends ReactContextBaseJavaModule {
   private static final String NSDocumentDirectoryPath = "NSDocumentDirectoryPath";
   private static final String NSExternalDirectoryPath = "NSExternalDirectoryPath";
   private static final String NSPicturesDirectoryPath = "NSPicturesDirectoryPath";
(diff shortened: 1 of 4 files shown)


android/src/main/java/com/rnfs/RNFSManager.java
```diff
@@ -332,6 +333,7 @@ private WritableMap makeErrorPayload(Exception ex) {
     final Map<String, Object> constants = new HashMap<>();
     constants.put(NSDocumentDirectory, 0);
     constants.put(NSDocumentDirectoryPath, this.getReactApplicationContext().getFilesDir().getAbsolutePath());
+    constants.put(NSTemporaryDirectoryPath, null)
```
9mm: @grabbou @johanneslumpe you are missing a `;` and this broke RNFS... @grabbou I will hunt you down and find you, and I will force you to eat Polish cabbage for 25 years for this riff-raff

grabbou: lol

PR #95, opened 2016-05-19
Update FS.common.js

grabbou:
I guess my PRs were not the best start of contributing. In a nutshell, it was a syntax error. Decided to remove it completely as we were not even using that (and I don't think it's desirable to have that console.log as a default listener)


grabbou:
But at least on the positive side we are testing the master in our release apps, so in case anything breaks - we will tell you :fire:

PR #96, opened 2016-05-20
MainBundlePath does not exist in android, changed to PicturesDirectoryPath

tristola:
Example in the Readme does not work in android.

Suggest using PicturesDirectoryPath, does it exist in both platforms?

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ public class MainActivity extends ReactActivity {
 var RNFS = require('react-native-fs');
 
 // get a list of files and directories in the main bundle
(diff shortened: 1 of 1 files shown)


johanneslumpe:
@tristola thanks. I do not have time to confirm whether the directory exists on both. Would be great if you were able to check and report back here.

tristola:
I haven't built apps yet for ios, so can only test for android. Learning now.

Seems that cocoapods needs more attributes with the latest version #101 - 
It was complaining about missing author, lisence and description.

I'll try to get the code example working with dependencies to be able to confirm that this works with ios too.

PR #97, opened 2016-05-20
Use React Native from node_modules

grabbou:
It wasn't probably a desired behaviour. Since maven contains outdated React, that's why it was throwing that the DeviceEventEmitter is not there etc. After changing it to use the node_modules one, it works again and all classes that are available in the currently installed React are working.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -25,10 +25,6 @@ android {
     }
 }
 
-repositories {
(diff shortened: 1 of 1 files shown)


grabbou:
PS. That fixes the `master` breaking with `RCTNativeAppEventEmitter` class is not found.

Issue #80, opened 2016-04-22
downloadFile() when app goes into the background


af:
I'm finding that for some large file downloads, the app going into the background (or the phone locking due to inactivity) is causing downloads via `downloadFile()` to fail with an error. Is there any workaround to support background downloads?

johanneslumpe:
@af this is a limitation of iOS - when your app goes into background any activity is stopped. What is the exact error you are receiving?

af:
I don't have the error handy unfortunately (I'm not at that computer at the moment). However, it looks like the ability to download in the background is now available with a new(ish) iOS API:

https://developer.apple.com/library/ios/documentation/iPhone/Conceptual/iPhoneOSProgrammingGuide/BackgroundExecution/BackgroundExecution.html
https://realm.io/news/gwendolyn-weston-ios-background-networking/

Not sure how much it would take to use that `NSURLSession` API for downloads instead of the existing `NSURLConnection` one (I'm an iOS noob). Does it look feasible with the existing RNFS design?

johanneslumpe:
@af It will require some code changes but without looking further into it I'd assume it is definitely possible. A downside I see is that it looks like iOS then will re-launch your app when the download completes which might not be a behavior people expect by default. (I might be reading that one wrong though).

af:
Cool! Glad to know it looks like it's possible. On the topic of re-launching, my reading of the docs suggests that's only the case when the `sessionSendsLaunchEvents` property is set to `YES`. That's just my interpretation though, could be incorrect

af:
@johanneslumpe I might be able to put together a PR for this later this week. Is it a reasonable approach to move all downloads to `NSURLSession`, or should there be an opt-in to this behaviour?

cjdell:
@af I think this should be opt-in behaviour. Android has a background download service of sorts so it might be possible to have an elegant cross platform abstraction. Others have suggested making `downloadFile` more capable so I'm debating having arguments passed as an `options` object so we can have optional things like `headers` as well. That would work well for this feature because we could have `{ background: true }` as an option.

af:
@cjdell Sorry, it took me longer than expected to be able to put together a PR for this! #100 contains an implementation of this feature, triggered by passing `{ background: true }` as an extra param to `downloadFile()` as you suggested.

johanneslumpe:
Closing as this is possible as of version `1.5.0`

PR #100, opened 2016-05-26
Add support for background downloads

af:
This was discussed in #80, including a suggestion to use the API implemented here. This PR only includes support for iOS background downloads.

1) Move downloads to use the newer `NSURLSession` APIs (supported in iOS 7 and later)
2) Add an extra `options` arg to `downloadFile()`. If this arg is `{ background: true }`, file downloads will persist when the app is backgrounded or the phone is put to sleep

Note that to use this setting, you need to request the "Background Fetch" capability for your app in XCode. I added a note about this in the Readme.

### DIFF
--- a/Downloader.h
+++ b/Downloader.h
@@ -16,7 +16,7 @@ typedef void (^ProgressCallback)(NSNumber*, NSNumber*);
 
 @end
 
-@interface Downloader : NSObject <NSURLConnectionDelegate>
+@interface Downloader : NSObject <NSURLSessionDelegate, NSURLSessionDownloadDelegate>
 
 - (void)downloadFile:(DownloadParams*)params;
 - (void)stopDownload;
--- a/Downloader.h
+++ b/Downloader.h
@@ -13,6 +13,7 @@ typedef void (^ProgressCallback)(NSNumber*, NSNumber*);
 @property (copy) ErrorCallback errorCallback;         // Something went wrong
 @property (copy) BeginCallback beginCallback;         // Download has started (headers received)
 @property (copy) ProgressCallback progressCallback;   // Download is progressing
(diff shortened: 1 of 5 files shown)


johanneslumpe:
Thoughts on this? cc @grabbou @cjdell

cjdell:
@af @grabbou @johanneslumpe Will merge this in soon. I want to sort out the argument structure so we can set request headers as well (like `uploadFiles`). Hope to sort this before the weekend.

af:
@cjdell Cool, thanks! FWIW Apple just approved an app using my fork into the App Store yesterday, so it's all good on that side of things

cjdell:
@af I've merged this in. I've tested it on my test app here:

https://github.com/cjdell/react-native-fs-test

There does seem to be an intermittent issue however:

image

Other than that it appears to be working.

af:
@cjdell Interesting, I haven't seen that message before. Is that on app startup or when you try and start a download?

cjdell:
@af I think it is happening when I press the button more than once. The first download seems to work, but later ones can fail. I'm also sometimes seeing in Xcode logs that a background download with the given URL is already in progress, however it never completes

cjdell:
@af Just pushed the added download test, forgot to push it.

af:
@cjdell Good catch, I'm guessing this is because I used the download url for the session identifier (argument to `backgroundSessionConfigurationWithIdentifier`). Apparently the best way to handle this is by passing a UUID instead so that it's globally unique across requests. I'll try and add a commit to fix this, should it be a separate PR or an addition to this one?

cjdell:
@af Thanks for looking into this. Please open a new PR as I've done quite a bit of refactoring since this one was opened. It will be much easier to merge in that way :-) Will close this PR now that it is reductant

PR #101, opened 2016-05-28
Fixed to work with cocoapods-1.0.0

tristola:
Added missing attributes to work with cocoapods-1.0.0


tristola:
Looks like the attributes should come from json, but do not work in my machine. Probaby just a problem with my machine.

PR #102, opened 2016-05-30
fix: workable path to external directory in android

Frosty92:
I tried to get access to the files/directories inside the external storage in Android like this: 

```
RNFS.readDir(RNFS.ExternalDirectoryPath)
        .then((result) => {
        console.log('GOT RESULT', result);
            .catch(console.log)
});
```

I got back an empty array, so I checked out how `ExternalDirectoryPath` was defined, which was: 
 `File externalDirectory = this.getReactApplicationContext().getExternalFilesDir(null);` 

The official documentation for [Android](https://developer.android.com/reference/android/os/Environment.html#getExternalStorageDirectory%28%29) has a different way of accessing the external directory, which is : `Environment.getExternalStorageDirectory()`. I tried this fix on Nexus 4, Moto E2, Nexus 6p and I am now able to get back an array of objects showing files and directories.

This pull request changes the definition of `ExternalDirectoryPath`.

### DIFF
--- a/android/src/main/java/com/rnfs/RNFSManager.java
+++ b/android/src/main/java/com/rnfs/RNFSManager.java
@@ -334,7 +334,7 @@ public class RNFSManager extends ReactContextBaseJavaModule {
     constants.put(NSDocumentDirectory, 0);
     constants.put(NSDocumentDirectoryPath, this.getReactApplicationContext().getFilesDir().getAbsolutePath());
     constants.put(NSTemporaryDirectoryPath, null);
-    File externalDirectory = this.getReactApplicationContext().getExternalFilesDir(null);
+    File externalDirectory = Environment.getExternalStorageDirectory();
     if (externalDirectory != null) {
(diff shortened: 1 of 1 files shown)


johanneslumpe:
This seems fine to me. Thoughts @cjdell ? I'm not sure as to whether `getReactApplicationContext` is an important part here and maybe shouldn't be ignored. (Despite the fact that something seems off)

cjdell:
@Frosty92 This is a tricky one. Do you know where previously `ExternalDirectoryPath` was pointing to before you made this change, and what it is after? Changing how this path is acquired will break things for existing users of the library who have stored persistent data.

superandrew213:
I am interested in this too.

Can't we just add a separate path ExternalStorageDirectoryPath to avoid breaking things? Android has an ExternalFilesDirectoryPath and ExternalStorageDirectoryPath.

superandrew213:
Forked it for my use case. Here is the [PR](https://github.com/johanneslumpe/react-native-fs/pull/121) if you want to add this.

PR #117, opened 2016-06-15
Return statusCode, headers for uploadFiles method

insraq:
1. Return statusCode, headers for uploadFiles method
2. Fix README for uploadFiles

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,16 +198,6 @@ var files = [
     filetype: 'audio/x-m4a'
   }
 ];
-// create an object of options
-var options = {
-  method: 'POST',
-  headers: {
-    'Accept': 'application/json',
-  },
-  fields: {
-    'hello': 'world',
-  }
-};
 
 var uploadBegin = (response) => {
   var jobId = response.jobId;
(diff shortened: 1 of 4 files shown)


cjdell:
@insraq Looks good, tested and working, thanks for this :-)

insraq:
Please note, this changes the response object in callback. So it's sort of a "breaking change"

johanneslumpe:
@cjdell @insraq This is not just sort of a breaking change, it is definitely one ;) This could be a non-breaking change if we still kept the `response` entry in the dictionary. Thoughts?

insraq:
@johanneslumpe Essentially you can. However, it would be confusing, as the old `response` is actually `body`. I will submit a new PR that includes `response` but won't mention it in the documentation. What do you think?

cjdell:
@insraq @johanneslumpe Shame on me for not spotting this, I have returned the key to being `response` as there is no (really) good reason for this to change. I'm now also capturing this behaviour in my test app.

cjdell:
@insraq @johanneslumpe Just realised in the docs that we're not actually describing the returned Promise object for some methods. I'll have to sort this out when I get another moment.

insraq:
@cjdell Can you take a look at this PR? https://github.com/johanneslumpe/react-native-fs/pull/118/files

I think `statusCode` `headers` `body` is much clearer. than `statusCode` `headers` `response`. I keep `response` there to keep it compatible.

cjdell:
@insraq Sorted, I've added them both in now. Did it manually as it was easier than dealing with the merge. :-)

insraq:
@cjdell Great. Thanks a lot!

PR #118, opened 2016-06-15
Keep it compatible

insraq:


cjdell:
@insraq Already done this, though not with both property names. Thought of doing that but I figured it would make the API more confusing. We can make breaking changes in V2. :-)

PR #123, opened 2016-06-22
ProgressDivider iOS fix & updates on every 1% of progress fix

locly:
This commit fixes two issues:
* On iOS when the progress divider was checked if it was 1, the check was being done on an `NSNumber` which meant it always failed. Converted it to an integer to fix.
* Originally the code used a `progressDivider <= 1` to indicate that it should send all calls across the bridge. It's quite a common use case to want to be updated on every full percentage though, so I have changed this to use a `progressDivider <= 0` as an indication that all progress calls should be sent across the bridge. This fix also changes the default progressDivider value to be 0 & updates the docs to reflect this

### DIFF
--- a/Downloader.m
+++ b/Downloader.m
@@ -68,7 +68,7 @@
   if ([_statusCode isEqualToNumber:[NSNumber numberWithInt:200]]) {
     _bytesWritten = @(totalBytesWritten);
 
-    if (_params.progressDivider <= 1) {
+    if (_params.progressDivider <= 0) {
       return _params.progressCallback(_contentLength, _bytesWritten);
     } else {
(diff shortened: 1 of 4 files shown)


johanneslumpe:
@cjdell Thoughts on this? This would be a breaking change.

locly:
Just as a side note @johanneslumpe  the progressDivider is only in master, not made it into a release yet, so any disruption should be minimal.

cjdell:
@johanneslumpe @locly I think we should accept this PR, it makes much more sense to have 0 mean everything so 1 can be used for each percentage point. Will give it a test now.

cjdell:
@johanneslumpe @locly Manually merged this in

PR #127, opened 2016-07-03
Add copyFile

superandrew213:


johanneslumpe:
Thanks! Sorry, was on vacation :) Why are you printing the stack trace before sending back the error on Android? Is that an Android best practice? Just wondering if there are any downsides to it in terms of performance (if it is not required)

superandrew213:
@johanneslumpe no problem :) 

I just followed the same structure you had for the other methods.

johanneslumpe:
@superandrew213 Ah ok - I didn't write any of the Android stuff. @cjdell can you shed some light on this?

superandrew213:
@johanneslumpe all other methods follow this structure. Maybe we can merge this PR then later evaluate performance issues since we have to change all other methods. What do you think?

johanneslumpe:
@superandrew213 sure, fine with me!

PR #131, opened 2016-07-07
Upload File Example documentation update.

asommer70:
Changed the uploadFiles options object to use **progressCallback** instead of **progress** because that is how it's named in FS.common.js.  This tripped me up until I dug into the source to figure out what the callback is actually named.

Note that the downloadFile option is named **progress** both the README example and the FS.common.js file.  Not sure if it's better to rename the uploadFiles option to match downloadFile option or not.

Thanks, this is a great extension to React Native.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,7 +221,7 @@ RNFS.uploadFiles({
       'hello': 'world',
     },
     begin: uploadBegin,
-    progress: uploadProgress
+    progressCallback: uploadProgress
   })
   .then((response) => {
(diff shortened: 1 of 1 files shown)


cjdell:
@asommer70 Good spot but this has been addressed in the new version:

https://github.com/johanneslumpe/react-native-fs/blob/master/FS.common.js

It was initially a mistake. Old names were put in for backwards compatibility but the new names were added to match up to the ones in `downloadFile`. We needed it to be consistent.

PR #142, opened 2016-07-26
[Android] Add read timeout

francisco-sanchez-molina:
Connection loose blocks download forever

### DIFF
--- a/android/src/main/java/com/rnfs/Downloader.java
+++ b/android/src/main/java/com/rnfs/Downloader.java
@@ -57,6 +57,7 @@ public class Downloader extends AsyncTask<DownloadParams, int[], DownloadResult>
       }
 
       connection.setConnectTimeout(5000);
+      connection.setReadTimeout(15000);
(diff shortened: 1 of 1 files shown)


johanneslumpe:
@francisco-sanchez-molina thanks for this. what does `setReadTimeout` do exactly? Does it time out if no data is read at all, or does it also timeout if some data is read and the connection subsequently fails in the middle of the download?

francisco-sanchez-molina:
Hi @johanneslumpe !

`setReadTimeout` does both. When the code is blocked on read (`while ((count = input.read(data)) != -1) {`),   `setReadTimeout` adds a watchdog for this read and that raises `java.net.SocketTimeoutException` if not completed at time. This allows close a download task that would otherwise be blocked infinitely.

https://docs.oracle.com/javase/7/docs/api/java/net/URLConnection.html#setReadTimeout(int)

johanneslumpe:
Ha, nice! Thanks!

PR #150, opened 2016-08-05
move-file fallback

corbt:
It turns out that the `renameTo` command used under the hood for the Android implementation of `moveFile` has some pretty serious [limitations](https://developer.android.com/reference/java/io/File.html#renameTo%28java.io.File%29). For example, it won't work to move something from the cache to the normal file system. This change still tries the `renameTo` command, but falls back to a copy-and-delete strategy if it fails.

### DIFF
--- a/android/src/main/java/com/rnfs/RNFSManager.java
+++ b/android/src/main/java/com/rnfs/RNFSManager.java
@@ -1,5 +1,6 @@
 package com.rnfs;
 
+import java.io.IOException;
 import java.util.Map;
 import java.util.HashMap;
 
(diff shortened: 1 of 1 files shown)


cjdell:
@corbt Thanks for this. My only feedback on this is that I think `copyFile` should be called internally for the sake of DRY, or maybe refactor both to call an internal private method. The implementation is almost identical.

corbt:
@cjdell refactored

cjdell:
@corbt You're obviously alert. Thanks :-)

PR #151, opened 2016-08-06
Make mkdir options optional

corbt:
This PR simplifies the code slightly by using a Javascript default value for `mkdir`'s `options` parameter. It also fixes a flow error that otherwise occurs when `mkdir` is called with only one argument.


cjdell:
@corbt Good spot, obviously my ES6 powers are weak :-)

PR #155, opened 2016-08-10
Add ScanFile method + Music and Downloads folder

alenoir:
- Add two folder : Music and Downloads
- Add scanFile method for indexing file with MediaScannerConnection


johanneslumpe:
Thanks! I I'm not sure if `scanFile` is the appropriate method name to use here, since it does not scan for all files, but only a subset.

Thoughts @cjdell ?

themakerman:
music and downloads folder @alenoir is this working with iOS now? Can i scan audio files stored locally in music and downloads folder for iOS devices using react-native-fs?

PR #160, opened 2016-08-17
Added length and offset parameters for reading files on iOS

macdabby:


cjdell:
@macdabby Thanks for this. It is my preference to make the API as Node-like as possible as a lot of JS devs know the Node FS API by heart. Node has a method simply called `read` which allows you to specify a length and offset. The only complication is it returns a Buffer which doesn't exist in RN.

macdabby:
I looked at that function but it looks like it doesn't have offset input, I guess because it's returning a buffer like you said? https://www.npmjs.com/package/read

Do you mean you want to make the function something like this:
var data = RNFS.read({
  bytes: 128,
  offset: 128,
});

PR #167, opened 2016-08-24
support https

Mr7Cat:
fail to download when url use https, fix it.

### DIFF
--- a/android/src/main/java/com/rnfs/Downloader.java
+++ b/android/src/main/java/com/rnfs/Downloader.java
@@ -10,6 +10,8 @@ import java.io.IOException;
 import java.net.URL;
 import java.net.URLConnection;
 import java.net.HttpURLConnection;
+import java.security.cert.CertificateException;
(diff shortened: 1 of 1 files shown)


johanneslumpe:
Just openly accepting all https certificates kind of defeats the purpose of using https in the first place in my opinion. Is there no other way to deal with this?

Also, is this an Android-only problem? (as there are no changes to the iOS code)

/cc @cjdell

Mr7Cat:
Just test with Android. I just find this method now...

cjdell:
@Mr7Cat If this is something people really need then we should probably make it available as an option, rather than just blindly accepting all self signed certs. It should be a relatively easy modification to make to this PR, but I would rather test it first.

johanneslumpe:
@cjdell I think the PR should also cover iOS, as this could just as well be an issue there too (I do know there are ways to accept self-signed certs etc. and that iOS does fail on those without further processing)

PR #168, opened 2016-08-24
Added Android getAsset method

mt81:
Hi, I have added the getAsset method in order to fetch the assets bundled within the .apk.

### DIFF
--- a/FS.common.js
+++ b/FS.common.js
@@ -188,6 +188,36 @@ var RNFS = {
     });
   },
 
+  readAsset(assetName: string, encodingOrOptions?: any): Promise<string> {
+    var options = {
+      encoding: 'utf8'
+    };
+
+    if (encodingOrOptions) {
+      if (typeof encodingOrOptions === 'string') {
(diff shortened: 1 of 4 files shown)


johanneslumpe:
Thanks! Would you mind refactoring the Javascript file so that the `readFile` and `readAsset` share common code? (which is 99%). I wonder if instead of always throwing on iOS, we should maybe try to read the file from the main application bundle instead since that's where assets are stored.

Thoughts on this?

cc @cjdell

mt81:
Absolutely! I can refactor the two methods and read files from the main bundle on iOS. Would it make sense to change the method name to something more generic? I was thinking readPackageResource or something similar... (Not good with names 😿  AHAH)

mt81:
it could also be a readFile option Eg. { fromAppBundle:bool }

benvium:
This is covered by https://github.com/johanneslumpe/react-native-fs/pull/83 now, so I'm going to close this PR. Thanks for your contribution!

PR #169, opened 2016-08-26
Update mkdir param in readme

asamiller:
The source looks like `mkdir` is expecting `filepath` and an `options` object but the docs say a boolean. Update the docs to show an object.


cjdell:
@asamiller Thanks for bringing this to my attention. They're actually both wrong, did a lot of refactoring recently and the key is actually `NSURLIsExcludedFromBackupKey`. Will fix it now...

PR #172, opened 2016-09-02
Added a much more convenient way to install this react-native-fs

shcheuk:
encourage people to use react native core link function


wafisher:
Does `react-native link` not work, then?

PR #173, opened 2016-09-02
Add a comment about RNFS.MainBundlePath

### LINKED ISSUE #133: Android RN 0.29.0 Example can not run
On RN 0.29.0, the getPackages function moved to MainApplication.java, so I edit it to:
```
......
@Override
    protected List<ReactPackage> getPackages() {
      return Arrays.<ReactPackage>asList(
          new MainReactPackage(),
          new RNFSPackage()
      );
    }
......
```
I edit the AndroidManifest.xml to add two permission
```
......
<uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
    <uses-permission android:name="android.permission.MOUNT_UNMOUNT_FILESYSTEMS"/>
......
```

then I follow the example to list file:

```
button1Pressed:function() {
        var RNFS = require('react-native-fs');
        // get a list of files and directories in the main bundle
        RNFS.readDir(RNFS.MainBundlePath)
            .then((result) => {
                console.log('GOT RESULT', result);

                // stat the first file
                return Promise.all([RNFS.stat(result[0].path), result[0].path]);
            })
            .then((statResult) => {
            if (statResult[0].isFile()) {
                // if we have a file, read it
                return RNFS.readFile(statResult[1], 'utf8');
                }

            return 'no file';
            })
            .then((contents) => {
            // log the file contents
                console.log(contents);
            })
        .catch((err) => {
            if ( err === null ) console.log('error is null');
            else console.log(err);
        });
    },
```

The project is successfully compiled and installed on my Samsung( Android 5.01) Mobile.

When button1Pressed is called, the error log is：
Error: Attempt to invoke virtual method 'char[] java.lang.String.toCharArray()' on a null object reference(…)

writeFile and readFile is OK

mixu:
Add a comment about RNFS.MainBundlePath not existing on Android.

The first example should work on Android, since it's probably the first thing anyone will try when getting started - I certainly was confused for a few minutes as to why the example didn't work.

Also fixes #133 - another person who was confused when getting started

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ public class MainApplication extends Application implements ReactApplication {
 var RNFS = require('react-native-fs');
 
 // get a list of files and directories in the main bundle
-RNFS.readDir(RNFS.MainBundlePath)
+RNFS.readDir(RNFS.MainBundlePath) // On Android, use "RNFS.DocumentDirectoryPath" (MainBundlePath is not defined)
   .then((result) => {
     console.log('GOT RESULT', result);
 


johanneslumpe:
Thanks!

PR #177, opened 2016-09-18
Remove trailing "/"

superandrew213:
Remove trailing "/" from TemporaryDirectoryPath to keep it consistent with the other paths.


johanneslumpe:
@superandrew213 This would be a breaking change for some users, no?

superandrew213:
@johanneslumpe yes it would be

itinance:
We shouldn't break compatibility here. Too dangerous.

PR #185, opened 2016-10-04
Check if callbacks are functions

corbt:
This fixes a Flow complaint that `NativeAppEventEmitter.addListener` requires a function, without having to add `beginCallback` and `progressCallback` to the options object (preferable because they're deprecated).


johanneslumpe:
Thanks!

PR #189, opened 2016-10-13
Add nil check on download completion callback

nmorel:
Hi, 

With version 1.5.1 on iOS, my app crash when I go to home during a download and back to the app with the following error `'NSInvalidArgumentException', reason: '*** -[__NSPlaceholderDictionary initWithObjects:forKeys:count:]: attempt to insert nil object from objects[3]`. 

I added a nil check to avoid the error.

### DIFF
--- a/RNFSManager.m
+++ b/RNFSManager.m
@@ -266,8 +266,10 @@ RCT_EXPORT_METHOD(downloadFile:(NSDictionary *)options
   params.progressDivider = progressDivider;
 
   params.completeCallback = ^(NSNumber* statusCode, NSNumber* bytesWritten) {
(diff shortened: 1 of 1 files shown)


johanneslumpe:
@nmorel do you know why this error occurs? Is it because the connection is aborted and thus there exists no status code?

nmorel:
I'm not sure. It only happens when the app lose the focus. Maybe iOS stops all the current connections.

johanneslumpe:
@nmorel yeah if your app goes into background mode and you do not request time to finish things (max ~ 10 minutes) i think, then connections are just being severed I think.

ToshKoevoets:
@johanneslumpe This is probably the case when te response body is empty. So a blank page

johanneslumpe:
I'll just go ahead and merge this, thanks!

PR #205, opened 2016-11-16
Add hash method for both iOS and Android

esamelson:
This PR adds a hash method to natively calculate file checksums using MD5 or SHA algorithms. Any feedback would be welcome!

### DIFF
--- a/FS.common.js
+++ b/FS.common.js
@@ -218,6 +218,10 @@ var RNFS = {
     });
   },
 
+  hash(filepath: string, algorithm: string): Promise<string> {
+    return RNFSManager.hash(filepath, algorithm);
+  },
+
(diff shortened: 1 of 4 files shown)


johanneslumpe:
This looks rather useful! :) How well has this been tested?

esamelson:
Well, it comes up with all of the proper checksums on the devices I've tested (various iOS devices and an Android 5 tablet), and it's quite performant for my use case -- calculating the checksum of a couple hundred ~10MB files at once. I have not tested it with significantly larger files (1GB+).

johanneslumpe:
Ok. Let's merge this in an go from there :) I don't have time to test this on my own but I trust you on this.

PR #214, opened 2016-11-29
Use existing NSURLSessionTaskDelegate method

aniastrzezek:
This PR fixes error not being thrown when download finishes with error due to eg. lost network connection.

### DIFF
--- a/Downloader.m
+++ b/Downloader.m
@@ -86,7 +86,7 @@
 
 - (void)URLSession:(NSURLSession *)session task:(NSURLSessionTask *)task didCompleteWithError:(NSError *)error
 {
-  return _params.errorCallback(error);
+  return error ? _params.errorCallback(error) : nil;
 }
 
 
--- a/Downloader.m
+++ b/Downloader.m
@@ -84,7 +84,7 @@
   return _params.completeCallback(_statusCode, _bytesWritten);
 }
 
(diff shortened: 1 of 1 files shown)


Thomas101:
This works a treat! I noticed that the commits come from a fork that's some 70 or so commits behind master, so I patched master locally (065acc400f6da3ce1cdf504ecc8be290a63bdb92) and it worked perfectly from there.

To test, start downloading file - place in aeroplane mode, the file download hangs indefinitely (no promise rejection or resolution). With the fix the promise rejects shortly after being placed offline

Thomas101:
Just noticed that if you cancel the download task this throws an error as the promise rejection gets called twice in 2.0.1-rc.2. The following fixes that behaviour

```js
- (void)URLSession:(NSURLSession *)session task:(NSURLSessionTask *)task didCompleteWithError:(NSError *)error
{
  if (error && error.code != -999) {
    _params.errorCallback(error);
  }
}
```

Happy to open a new PR for this if it helps, we have a fork over at https://github.com/locly/react-native-fs

aniastrzezek:
@Thomas101 Thanks for your comments!

Fork is based on the last released version ([1.5.1](https://github.com/johanneslumpe/react-native-fs/releases/tag/1.5.1)). I'll take a look on task canceling issue!

PR #219, opened 2016-12-14
iOS stopDownload Race Condition

nbolender:
When using RNFS's `downloadFile` functionality, the user's javascript component may call `stopDownload` after the file has actually completed downloading on iOS. Because of a delay between communicating between javascript and iOS, the javascript component may not have executed its `completeCallback` despite it being queued on the native side. This means that the javascript component "thinks" that the download is still occurring, so it may call stopDownload.

But, because the download has completed, the success and complete callbacks have been discarded by React Native. **This triggers an error that reads "Callback with id #: undefined.undefined() not found".**

This PR includes a simple check to only cancel the download if it is still downloading, thus avoiding the race condition that causes the above error.

### DIFF
--- a/Downloader.m
+++ b/Downloader.m
@@ -107,15 +107,17 @@
 
 - (void)stopDownload
 {
-  [_task cancel];
+  if (_task.state == NSURLSessionTaskStateRunning) {
+    [_task cancel];
 
-  NSError *error = [NSError errorWithDomain:@"RNFS"
-                                       code:@"Aborted"
-                                   userInfo:@{
-                                     NSLocalizedDescriptionKey: @"Download has been aborted"
-                                   }];
+    NSError *error = [NSError errorWithDomain:@"RNFS"
+                                         code:@"Aborted"
+                                     userInfo:@{
+                                       NSLocalizedDescriptionKey: @"Download has been aborted"
+                                     }];
 
-  return _params.errorCallback(error);
+    return _params.errorCallback(error);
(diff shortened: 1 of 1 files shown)


johanneslumpe:
@nbolender Since I haven't implemented the download functionality, I just want to make sure: If a download is already completed and then someone aborts, is the download promise still going to be resolved?

nbolender:
@johanneslumpe In that case the download promise has already been resolved, so calling the error callback won't work anyway. Once you resolve or reject the promise, React Native gets rid of both the success and failure callbacks, so calling either triggers an error.

It is an edge case, but it is one I was seeing frequently in my application with ~60 concurrent downloads. Here is the timeline of the edge case:

1. JS component starts download & native component creates download task
2. Download completes; native component resolves the promise which queues the success callback to be run on the JS thread
3. Something triggers the JS component to want to cancel the download (such as an unmount)
4. JS component still has not executed the success callback that has been queued on the native side, so it calls stopDownload() instead of doing nothing
5. Native component creates an error and tries to reject the promise, which queues the error callback to be run on the JS thread
6. The success callback is finally run on the JS component
7. The error callback reaches React Native's MessageQueue.js, but the callback has already been freed from memory when the success callback was run
8. A React Native error is triggered with the description "Callback with id #: undefined.undefined() not found"

johanneslumpe:
@nbolender Makes sense, thanks!

PR #223, opened 2016-12-23
[Breaking] 0.40 Add namespaced import React/* to headers

ptmt:
Related: https://github.com/facebook/react-native/commit/e1577df1fd70049ce7f288f91f6e2b18d512ff4d

Also requires updates in Example. So I guess, I left it here for those, who for some reasons want to update to RN@0.40.


adrienbataille:
@ptmt Hi, can you please merge this PR? 0.40 is out already.

jhampton:
Bumping the above PR

avioli:
This is a weird version change. Why bump major version _and_ the RC? Why include a version bump commit at all? My understanding is that pull requests should be from a branch that doesn't affect the version, thus `master` is not the best use case.

React Native _advises_ imports to use angle brackets: `#import <React/RCT...>`, not quotes `#import "React/RCT..."`. Both work though, just a matter of prioritisation - double quotes first checks local directory, and if it doesn't find a match moves to system headers paths, whereas the angle brackets skips directly to the system paths.

ptmt:
@avioli You're right, as I said it was rush update (I forked and updated like 6-7 plugins). Now, when 0.40 is released it's time to make a proper pull request.

PR #227, opened 2016-12-29
Fix ctime/utime attributes on .stat()

### LINKED ISSUE #222: File timestamps in .stat()?
Hi,
Is there some technical reason why .stat() isn't returning created/updated timestamps? Would you accept a PR if there isn't something major blocking it?

aforty:
This fixes the .stat() results so that ctime/utime come through as expected. Looks like it was just an oversight. Fixes #222.


johanneslumpe:
Ha, look at that. Thanks!

PR #235, opened 2017-01-11
Update README.md

baskan:
update readme npm install instructions acording to #223 breaking changes. 
new users should be informed.


johanneslumpe:
Thanks!

PR #238, opened 2017-01-15
Closes #46 - 'RCTBridgeModule.h' file not found error

### LINKED ISSUE #46: Build Failed on iOS
There is an error: 'RCTBridgeModule.h' file not found.

iSimar:
This PR for the fix of issue: https://github.com/johanneslumpe/react-native-fs/issues/46

### DIFF
--- a/RNFSManager.h
+++ b/RNFSManager.h
@@ -6,8 +6,8 @@
 //  Copyright (c) 2015 Johannes Lumpe. All rights reserved.
 //
 
-#import <React/RCTBridgeModule.h>
-#import <React/RCTLog.h>
+#import "RCTBridgeModule.h"
+#import "RCTLog.h"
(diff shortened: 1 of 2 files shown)


cjdell:
This is more of a workaround than a fix. If you create a new app with the CLI tool it will use arrow brackets. We should find out why arrow brackets isn't working in this case.

iSimar:
@cjdell hmm...interesting. you're right.

clarkie:
@cjdell which CLI tool are you referencing?

itinance:
This PR would break compatibility with ReactNative >= 0.40 (and in fact it will undo the changes that where necessary when 0.40 came out). 

If you are below RN 0.40, consider to use an older release of react-native-fs please.

PR #245, opened 2017-01-30
[iOS] Cancelling download causes promise rejection to be called twice

Thomas101:
If you cancel a download task the promise rejection is called twice by the delegate. This checks for -999 status code which indicates the download has been cancelled


Thomas101:
Hi, is this something you'd be interested in merging in? Thanks

PR #246, opened 2017-02-01
Add pathForGroup method

alexey-pkv:
When two applications share the same group, a common folder can be accessed to share data between this two applications. This method will return the path to a shared group folder by the group's name. See [containerURLForSecurityApplicationGroupIdentifier](https://developer.apple.com/reference/foundation/nsfilemanager/1412643-containerurlforsecurityapplicati) and [About App Sandbox](https://developer.apple.com/library/content/documentation/Security/Conceptual/AppSandboxDesignGuide/AboutAppSandbox/AboutAppSandbox.html#//apple_ref/doc/uid/TP40
011183) for reference.


android/src/main/java/com/rnfs/RNFSManager.java
```diff
@@ -562,7 +562,12 @@ public void stopDownload(int jobId) {
 
   @ReactMethod
   public void pathForBundle(String bundleNamed, Promise promise) {
-    // TODO: Not sure what equilivent would be?
+    // TODO: Not sure what equivalent would be?
+  }
+
+  @ReactMethod
+  public void pathForGroup(String bundleNamed, Promise promise) {
+    // TODO: Not sure what equivalent would be?
```
itinance: I would suggest to reject this promise with a meaningful error message

itinance:
Nice catch! Thanks for contributing. Do you might adding a section to the Readme describing the purpose of this method (with special warning that this is iOS only)?

alexey-pkv:
@itinance Yes of course. I will look into it next week.

alexey-pkv:
Done, please review.

PR #254, opened 2017-02-14
Made new thread for other native processes

codesinghanoop:
i wanted to process other native functions, but downloading process was blocking my other native process such as get data from local storage. When downloading process completes then the control goes to remaining native processes till than it keeps it in a queue.

### DIFF
--- a/android/src/main/java/com/rnfs/Downloader.java
+++ b/android/src/main/java/com/rnfs/Downloader.java
@@ -22,20 +22,23 @@ import com.facebook.react.bridge.ReadableMapKeySetIterator;
 public class Downloader extends AsyncTask<DownloadParams, int[], DownloadResult> {
   private DownloadParams mParam;
   private AtomicBoolean mAbort = new AtomicBoolean(false);
+  DownloadResult res;
 
   protected DownloadResult doInBackground(DownloadParams... params) {
     mParam = params[0];
+    res = new DownloadResult();
 
-    DownloadResult res = new DownloadResult();
-
-    try {
-      this.download(mParam, res);
-      mParam.onTaskCompleted.onTaskCompleted(res);
-    } catch (Exception ex) {
-      res.exception = ex;
-      mParam.onTaskCompleted.onTaskCompleted(res);
-      return res;
-    }
+    new Thread(new Runnable() {
+      public void run() {
+        try {
+          download(mParam, res);
+          mParam.onTaskCompleted.onTaskCompleted(res);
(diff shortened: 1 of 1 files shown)


br4nnigan:
is it possible that currently downloads block each other in that manner? That means only one download at a time. I have this impression but will investigate further.

itinance:
Sounds very usefully! Any Java-Head out there to help double check this PR? Maybe @benvium ?

itinance:
I will merge this in upcoming days, together with a solution for iOS also by my own. Thought i could do this much faster but it will take some days

br4nnigan:
\o/

a-koka:
@itinance, any update on this for iOS? I'd love to have parallel downloads on iOS. Happy to help out if you haven't made much progress.

a-koka:
@codesinghanoop What about using a `ThreadPoolExecutor` instead of spawning a new `Thread` per download? I'm happy to provide a PR where the pool can be configured or perhaps just limit it to 8 simultaneous downloads.

What do you think?

itinance:
@a-koka unfortunately i haven't much progress on this for iOS. Any help with PR's are much appreciated :) Thank you!

a-koka:
@codesinghanoop Can you help me understand why long tasks will definitely fail? I may be missing something here.

I have a lot of experience with `ThreadPoolExecutor`. It can be configured to not have idle threads so there should not be an issue with idle threads. Plus the cost of spinning new threads for each download will be removed including a faster download as for the case explained below.

For apps like mine that need to download several images or other assets upon installation (which I assume is a common use case), devices end up spinning 40-50 (in my case) threads and parallel downloads at the same time. This causes slowness due to a high number of threads and multiple connections to download. Having a thread pool would greatly help with this.

Hope I'm not missing something here. Happy to submit a PR for review.

a-koka:
@itinance Sounds good. It's not at the top of my list but I would love to work on this and submit a PR in the next few weeks.

codesinghanoop:
Long task will fail because the main thread do not know when all the tasks are completed and it can start to process the result. It could be a complicated implementation for this lib. There are edge cases in this implementation.

codesinghanoop:
There are very few disadvantages of ThreadPoolExecutor however, I would love to see how you handle the edge cases. Please let me know if you need my help.

pudility:
@codesinghanoop This does not seem to be working for me - maybe I am doing something wrong but when I download a file all async functions still halt until after the file has been downloaded. What can I do to fix this?

PR #256, opened 2017-02-16
Fix doc of stopDownlaod

cpunion:
It returns void, not Promise<void>


benvium:
Thanks, you are right.

PR #264, opened 2017-03-01
fix adding RNFS with CocoaPods

futuun:
currently if you try to add this package with CocoaPods you will get:
```
[!] Invalid `RNFS.podspec` file: Malformed version number string v2.1.0-rc.1.

 #  from {...}/node_modules/react-native-fs/RNFS.podspec:13
 #  -------------------------------------------
 #    s.platform        = :ios, "7.0"
 >    s.source          = { :git => "https://github.com/johanneslumpe/react-native-fs", :tag => "#{s.version}" }
 #    s.source_files    = '*.{h,m}'
 #  -------------------------------------------
```

because there is 'v' in package.json version. This pr will fix that


Myrenkar:
@johanneslumpe Could you please look at this?

Issue #253, opened 2017-02-10
Is there anyway to include the creation date of the file?


crysfel:
Hi folks!

I'm wondering if is possible to add the creation date of the file to the `readDir` method. It would be nice to include this information on the list of objects, currently we have `name`, `path` and `size` which is very useful, but I think the creation date should be there as well.

Any plans on adding this feature? Is it even possible to get this information from the file system on iOS and Android devices?

Regards

Ignigena:
Needing this functionality myself and just submitted a PR for testing.

itinance:
waiting for a [small change](https://github.com/itinance/react-native-fs/pull/265#pullrequestreview-34813768) [or discussion], then i will merge quickly

itinance:
As with stat() there is now "ctime" and "mtime" available also in readDir(). Thanks to #283

PR #265, opened 2017-03-08
Add created and modified dates to `ReadDirItem`

Ignigena:
Found myself needing this functionality when scanning a directory for file contents using `readDir`. The `created` date is only available on iOS but the `modified` date is available for both platforms.

Related issue: #253

Disclaimer: I'm not an Android developer and only a recent React Native developer so if there's a better way to accomplish this please point it out!

### DIFF
--- a/FS.common.js
+++ b/FS.common.js
@@ -31,6 +31,8 @@ type MkdirOptions = {
 };
 
 type ReadDirItem = {
+  created: date;     // The creation date of the file (iOS only)
+  modified: date;     // The last modified date of the file
   name: string;     // The name of the item
   path: string;     // The absolute path to the item
   size: string;     // Size in bytes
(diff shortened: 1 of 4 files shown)


FS.common.js
```diff
@@ -159,6 +161,8 @@ function readFileGeneric(filepath: string, encodingOrOptions:?string, command: F
 function readDirGeneric(dirpath: string, command: Function) {
   return command(normalizeFilePath(dirpath)).then(files => {
     return files.map(file => ({
+      created: file.created && new Date(file.created),
```
itinance: Thank you for contributing! I think it is not a good idea to deliver the current date on Android (as file.created will not be available). We should instead deliver "null" here to indicate that this attribute simply not exists. Can you change this that way? i will merge then. thx :)

itinance:
Thank you for contributing! See my comments please :)

Ignigena:
Re-opened under #283 ... I deleted my fork because it seemed the project had been abandoned with a deprecation notice. Made the requested changes, let me know if anything else needs changing :)

crysfel:
@itinance is this project still alive? It was my understanding that it was abandoned o.O I believe this project is quite popular and really useful :)

Regards

itinance:
@crysfel yes, it is back under active maintainance

PR #270, opened 2017-03-19
add read android raw resource file

fov42550564:
add readFileRaw/copyFileRaw/existsRaw for android

### DIFF
--- a/android/src/main/java/com/rnfs/RNFSManager.java
+++ b/android/src/main/java/com/rnfs/RNFSManager.java
@@ -156,6 +156,34 @@ public class RNFSManager extends ReactContextBaseJavaModule {
   }
 
   @ReactMethod
+  public void readFileRaw(String filename, Promise promise) {
+    InputStream stream = null;
+    try {
+      int res = getReactApplicationContext().getResources().getIdentifier(filename, "raw", getReactApplicationContext().getPackageName());
+      stream = getReactApplicationContext().getResources().openRawResource(res);
+      if (stream == null) {
+        reject(promise, filename, new Exception("Failed to open file"));
+        return;
+      }
+
+      byte[] buffer = new byte[stream.available()];
+      stream.read(buffer);
+      String base64Content = Base64.encodeToString(buffer, Base64.NO_WRAP);
+      promise.resolve(base64Content);;
+    } catch (Exception ex) {
+      ex.printStackTrace();
+      reject(promise, filename, ex);
+    } finally {
+      if (stream != null) {
+        try {
+          stream.close();
+        } catch (IOException ignored) {
+        }
+      }
+    }
+  }
+
+  @ReactMethod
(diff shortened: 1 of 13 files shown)


android/src/main/java/com/rnfs/RNFSManager.java
```diff
@@ -324,6 +355,20 @@ public void copyFileAssets(String assetPath, String destination, Promise promise
   }
 
   @ReactMethod
+  public void copyFileRes(String filename, String destination, Promise promise) {
+    String prefix = filename.substring(filename.lastIndexOf(".")+1);
+    String name = filename.substring(0, filename.lastIndexOf("."));
+    Boolean isImage = prefix.equals("png") || prefix.equals("jpg") || prefix.equals("jpeg") || prefix.equals("bmp") || prefix.equals("gif") || prefix.equals("webp") || prefix.equals("psd") || prefix.equals("svg") || prefix.equals("tiff");
```
itinance: This code exists 3 times in the file now. In order to reduce duplicated code it would be better to have an own static method here called "isImage" that would be called from all 3 occurrences where this detection is needed

android/src/main/java/com/rnfs/RNFSManager.java
```diff
@@ -156,6 +156,37 @@ public void readFileAssets(String filepath, Promise promise) {
   }
 
   @ReactMethod
+  public void readFileRes(String filename, Promise promise) {
+    String prefix = filename.substring(filename.lastIndexOf(".")+1);
+    String name = filename.substring(0, filename.lastIndexOf("."));
+    Boolean isImage = prefix.equals("png") || prefix.equals("jpg") || prefix.equals("jpeg") || prefix.equals("bmp") || prefix.equals("gif") || prefix.equals("webp") || prefix.equals("psd") || prefix.equals("svg") || prefix.equals("tiff");
+    InputStream stream = null;
+    try {
+      int res = getReactApplicationContext().getResources().getIdentifier(name, isImage ? "drawable" : "raw", getReactApplicationContext().getPackageName());
```
itinance: until this line and everything before "try" and after "catch" is the same in readFileRes, copyFileRes and existsRes. I understand the purpose, but wouldn't it make sense to isolate all this boilerplate in a separated private method and give it a callback what to do with it finally (reading, copying, existance-check) in order to reduce duplicated code and make it more clean

itinance:
Thank you for contributing! I made some comments to consider in order to make this code less duplicated and more clean. Can you improve it a bit? Will merge then. Anyways, very usefull contribution!

heuism:
hi guys, this PR have some conflict with the master. Can I ask if this one is already resolved? or can I copy the raw file in raw folder without using this? Thanks

kytwb:
What's the status here?

nihaux:
ping ? interested in this too

itinance:
Closed in favour of #645

PR #271, opened 2017-03-21
Update the android SDK version to 25

markusguenther:
When using the current android studio with the current SDKs you have the problem that your app is on version 25 and you can not create a APK without adjusting the buildToolsVersion in the build.gradle

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -11,8 +11,8 @@ buildscript {
 apply plugin: 'com.android.library'
 
 android {
-    compileSdkVersion 23
-    buildToolsVersion "23.0.1"
+    compileSdkVersion 25
+    buildToolsVersion "25.0.0"
 
     defaultConfig {
         minSdkVersion 16


itinance:
I think we should leave it on par with react native itself. More over, lots of applications outside are still on SDK 23 and also all the new ones that are created with "react-native init".

markusguenther:
I know that react-native is on version 23. But if you're using android studio. The IDE often treat you to update the IDE and later on Gradle. Then you are somehow at the point where you need to adjust some libraries you use to version 25. Or whatever your SDK version will be.

itinance:
Hm.... the folks from [react-native-camera](https://github.com/lwansbrough/react-native-camera) also refer to update buildtools to 25

markusguenther:
@itinance Did this PR in some other packages and common sense was that they stay on version 23 as long as react-native does.

itinance:
Thx for clarifying. @markusguenther. Lets keep this open until RN will update itself its build tools.

markusguenther:
I actually add this to all projects. So I don`t care for this anymore. But hope that RN updates this soon.

```
subprojects {
    afterEvaluate {project ->
        if (project.hasProperty("android")) {
            android {
                compileSdkVersion 25
                buildToolsVersion '25.0.0'
            }
        }
    }
}
```

This code is in my root build.gradle

itinance:
Yes i did this also in all projects :)
But with next release we will already be at 25.

PR #283, opened 2017-04-26
Add created and modified dates to `ReadDirItem`

Ignigena:
Originally from #265 but I had deleted our fork thinking the project was abandoned.

`created` attribute is `null` on Android with these changes as requested!

### DIFF
--- a/FS.common.js
+++ b/FS.common.js
@@ -31,6 +31,8 @@ type MkdirOptions = {
 };
 
 type ReadDirItem = {
+  created: date;     // The creation date of the file (iOS only)
+  modified: date;     // The last modified date of the file
   name: string;     // The name of the item
   path: string;     // The absolute path to the item
   size: string;     // Size in bytes
@@ -159,6 +161,8 @@ function readFileGeneric(filepath: string, encodingOrOptions:?string, command: F
 function readDirGeneric(dirpath: string, command: Function) {
   return command(normalizeFilePath(dirpath)).then(files => {
     return files.map(file => ({
+      created: file.created && new Date(file.created) || null,
+      modified: new Date(file.modified),
       name: file.name,
       path: file.path,
       size: file.size,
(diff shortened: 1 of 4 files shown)


itinance:
There is a strange issue. I got the error message

> RCTJSONStringify() encountered the following error: Invalid type in JSON write (__NSTaggedDate)

 when readDir() of MainBundle-Directory. Turns also out, that some items are not really NSStrings that can't converted into JSON-valid Values:

When i comment out these [both lines](https://github.com/itinance/react-native-fs/pull/265/commits/89dbe482e9af85237339e588b2c7fb9981ed6d09#diff-5fd3f3225a9de07b96ff3fcd573c0d27R55) out in RNFSManager, it works pretty fine again.

It seems that the JsonSerializer is not happy with this kinde of NSTaggedDate. But i wonder why this works for you?

itinance:
Many sources are saying "you can't convert NSDate to JSON, use NSString-presentation". Maybe we should work with unix timestamp here as it is done also in the [stat-function](https://github.com/itinance/react-native-fs/blob/89dbe482e9af85237339e588b2c7fb9981ed6d09/FS.common.js#L243)?

itinance:
@Ignigena i did a small [change](https://github.com/itinance/react-native-fs/commit/a5ee52f5c8b1ea024ac6a95978afb40fc00679dd) in order to fix this json-serialization-error. I found out that the stat-method had a similar issue and used its approach here. Moreover i renamed the properties "created" and "modified" to "ctime" respectively "mtime" in order to be compatible with the stat()-method.
Maybe you need to change your app-code this way now, sorry for inconvinience. But i think that a common naming scheme across all methods and parameters should be priotized :)

vincon09:
ctime and mtime are giving same timestamps ? pls help

vincon09:
ctime and mtime are giving same timestamps ? pls help

Issue #250, opened 2017-02-07
Support for writing/updating data at any given position in the file


IjzerenHein:
Hi, first of all, awesome library, keep up the good works! 👍👍

In our application we have an additional need for updating files (random access writes, at any given position in the file, not just rewriting and appending). We currently have this working on a custom fork and it's been battle-tested on iOS and Android for about half a year now. We now want to push this feature into `react-native-fs`, but I first want to verify a proper API for this with you. I know that you're keen on keeping the API as similar as possible to node-js, and I completely share that sentiment (although our current API is slightly different atm).

So, looking at how node-js does it, what would you think of the following API for updating files?

```
RNFS.write(filepath: string, buffer: Uint8Array, offset: integer, length: integer, position: integer): Promise<void>

Write the buffer to filepath. Offset determines the part of the buffer to be written, and length is an integer specifying the number of bytes to write. Position refers to the offset from the beginning of the file where this data should be written. If typeof position !== 'number', the data will be written at the current position.
```

Do you consider this a good API for random access writes?

Looking forward to your feedback!

Cheers,
Hein

itinance:
Hi @IjzerenHein, are you still available for contribution? Since some days this library is back maintained again (i am the new maintainer) and your proposed solution seems to fit to other users requests. Your suggested API looks fine to me!

IjzerenHein:
That's awesome to hear! And really glad to see this library is being maintained again :)
I'l create a PR for this and let you know.

IjzerenHein:
I've added the PR #288 
Let me know once the PR is accepted and has been released, so I can upgrade my project.
Cheers, Hein

PR #288, opened 2017-04-29
Added write function for performing random access writes

IjzerenHein:
This pull requests adds support for random access writes, as discussed in #250.
It follows the official node-js FS api, signature:
https://nodejs.org/api/fs.html#fs_fs_write_fd_string_position_encoding_callback

### `write(filepath: string, contents: string, position?: number, encoding?: string): Promise<void>`

Write the `contents` to `filepath` at the given random access position. When `position` is `undefined` or `-1` the contents is appended to the end of the file. `encoding` can be one of `utf8` (default), `ascii`, `base64`.

### DIFF
--- a/FS.common.js
+++ b/FS.common.js
@@ -339,6 +339,38 @@ var RNFS = {
     return RNFSManager.appendFile(normalizeFilePath(filepath), b64);
   },
 
+  write(filepath: string, contents: string, position?: number, encodingOrOptions?: any): Promise<void> {
+    var b64;
+
+    var options = {
+      encoding: 'utf8'
+    };
+
+    if (encodingOrOptions) {
+      if (typeof encodingOrOptions === 'string') {
+        options.encoding = encodingOrOptions;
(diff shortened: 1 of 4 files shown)


itinance:
Hey @IjzerenHein, thanks for contributing! Good work!

One thing: `write()` has a similar signature (and purpose) like `writeFile()`. What if we would extend the already existing method "writeFile" with the optional parameter "position" and handle this case then  with your implementation instead? This would keep the structure of RNFS more clean and no one would be confused when having a method "writeFile" and also "write", which have the same signature except for "position".

What do you think?

IjzerenHein:
@itinance that's a great idea. I initially had the same idea but I thought the original author wanted to adhere to the nodejs FS API as much as possible.
Go ahead make the change, or do you want me to make it?

itinance:
Good hint, @IjzerenHein! I follow the argumentation to leave it close to the NodeJS-API as i read it here in more issues already that the people like to follow it.

Thank you, i will merge now.

PR #290, opened 2017-05-01
Fix flowtype of date object in ReadDirItem

### LINKED ISSUE #289: Flowtype nonexistent `date` type
Flowtype not correct on b139b5d58b309ae054995400187ee3632188d732

```sh
 34:   created: date;     // The creation date of the file (iOS only)
                ^^^^ identifier `date`. Could not resolve name
```

Will put up a PR shortly.

wschurman:
### Summary

On react-native projects that use flow (on by default), this started showing up as an invalid type.

This commit also upgrades flow bin to match .flowconfig version and adds a npm script to run flow check.

Resolves #289

### Test plan

1. Run `npm run flow`, no longer see error for undefined `date` type.
2. Import into another project locally, no longer see that error.

### DIFF
--- a/FS.common.js
+++ b/FS.common.js
@@ -31,8 +31,8 @@ type MkdirOptions = {
 };
 
 type ReadDirItem = {
-  created: date;     // The creation date of the file (iOS only)
-  modified: date;     // The last modified date of the file
+  created: ?Date;    // The creation date of the file (iOS only)
+  modified: Date;   // The last modified date of the file
   name: string;     // The name of the item
   path: string;     // The absolute path to the item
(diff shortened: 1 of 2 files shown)


itinance:
Thanks for contributing! Please consider that created/modified was renamed to mtime/ctime in order to be compatible with result of stat()-function (although it wasnt reflected right in ReadDirItem). I changed this in ReadDirItem after the merge.

lukaleli:
When this change will be released?

itinance:
I am currently adding some further methods and will release today apparently

lukaleli:
Thanks! Looking forward to it.

PR #291, opened 2017-05-02
[tvOS] Add target

JulienKode:
Add tvOS target to use react-native-fs on tvOS


itinance:
Thank you very much for contribution!

JulienKode:
Awesome project 👍 @itinance

itinance:
Thx, but i won't adorn myself with borrowed plume. It was originated by @johanneslumpe, who did the most work here :)

PR #301, opened 2017-06-01
added additional download input parameters readTimeout and connectionTimeout

drunksaint:
done for android right now. didn't seem like ios had the same timeouts set.


itinance:
Thank you very much for contributing! I will look for equivalent iOS-Settings in upcoming days

PR #302, opened 2017-06-09
pod file path error

colorfulberry:
pod file path error


itinance:
Nice catch! Thank you very much!

PR #305, opened 2017-06-14
Remove createJSModules from package

javache:
createJSModules is now no longer required https://github.com/facebook/react-native/commit/ce6fb337a146e6f261f2afb564aa19363774a7a8

This PR is currently only compatible with RN master.


itinance:
Thank you vey much!

itinance:
@javache I'm also curious to know if facebook is using this module :)

javache:
Yeah, we're using this for one of our internal apps.

chirag04:
this is a breaking change, right ?

javache:
Correct. I'm not aware of a better way to get rid of this API without making it a breaking change.

PR #310, opened 2017-06-30
add `touch` method for changing file timestamps

formatlos:
I added a new method for changing the creation and modification timestamp of files.


itinance:
Great! Thank you very much!

Issue #153, opened 2016-08-08
Ability to read byte range [Feature Request]


macdabby:
In some instances, uploading a file in chunks is more reliable - an uploaded file can be uploaded partially then continued later if the connection is lost. In order to do this, it would be great if a file stream could load just a segment of a file starting at a particular offset and loading a specific number of bytes.

johanneslumpe:
@macdabby I agree that this would be a nice addition. Right now I cannot devote any time to this though. Feel like submitting a PR to add this functionality for both iOS and Android?

macdabby:
Yeah, I was thinking about it :) I'm still very new to RN so it might take me a while to figure some of it out.

macdabby:
Hey, what do you know, this is pretty easy :D I don't have my android environment setup yet, but I'll try to get that working next. Here is a PR for iOS using the same methods with additional params:
https://github.com/johanneslumpe/react-native-fs/pull/160

faceyspacey:
+1

PR #314, opened 2017-07-16
read method similar to the node.js fs.read method (iOS and Android)

simitii:
I need to read files piece by piece, in other words chunk by chunk, in my project. So current methods such as `readFile` method DO NOT meet my requirements. I wrote a `read` method for my requirement. I think other developers may need similar functionality.

`read(filepath: string, length = 0, position = 0, encodingOrOptions?: any): Promise<string>`

I tried to use a method signature similar to the fs.read method of Node.js
But I think `buffer` and `offset` params are not suitable for our case. 
[reference to fs.read Node.js](https://nodejs.org/dist/latest-v6.x/docs/api/fs.html#fs_fs_read_fd_buffer_offset_length_position_callback)

I took the iOS implementation from an old pull request.
So credits for iOS implementation goes to @macdabby 

#160
#153

### DIFF
--- a/FS.common.js
+++ b/FS.common.js
@@ -260,6 +260,36 @@ var RNFS = {
     return readFileGeneric(filepath, encodingOrOptions, RNFSManager.readFile);
   },
 
+  read(filepath: string, length = 0, position = 0, encodingOrOptions?: any): Promise<string> {
+  	var options = {
+      encoding: 'utf8'
+    };
+
+    if (encodingOrOptions) {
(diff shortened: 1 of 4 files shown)


macdabby:
👍  nice!

itinance:
WOW! Thank you very much. 

For now pls use master. I will create a new release in the upcoming week

Shannanigans:
Hi guys

When can we expect a release for the read function? I require this kind of functionality to handle large file uploads.

Thanks for the great work.

PR #318, opened 2017-07-19
Update `compileSdkVersion` to 26 and `buildToolsVersion` to 26.0.0

joshuapinter:
Minimum required for up-to-date apps is 25. This saves people from getting the following error and having to fix it manually.

screenshot 2017-07-19 12 59 48

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -11,8 +11,8 @@ buildscript {
 apply plugin: 'com.android.library'
 
 android {
-    compileSdkVersion 23
-    buildToolsVersion "23.0.1"
+    compileSdkVersion 26
+    buildToolsVersion "26.0.0"
 
     defaultConfig {
         minSdkVersion 16


itinance:
Thank you! There is already a PR existing concerning 25. 
The common opinion across several RN-libraries and their maintainers was to stay with the setting that RN comes with itself.

One question:  will this force all people to install SDK 26 or can they also compile with 24 or 25?

joshuapinter:
Yes, probably a good rule of thumb to stick with what RN ships with.

From my experience, you will need to have the SDK Build Tools installed for Version 26 (or whatever version you set). Not sure if you need the entire SDK but definitely the build tools for that version.

I'd be happy with Version 25 just to avoid these errors. They're especially bad when you use `yarn` because yarn wipes out the package in `node_modules` everytime you do a `yarn install`.

itinance:
Okay, thx for clarifying. I just merged PR with build version 25

joshuapinter:
Awesome, thanks! Will close this, then.

itinance:
You're welcome. And thank you very much for clarifying!

joshuapinter:
👍

PR #319, opened 2017-07-20
Add `createJSModules()` back to RNFSPackage.

joshuapinter:
This causes an error for me on react-native 0.46.4.

Fixes #317.

PR #324, opened 2017-08-02
removed createJSModules method (RN 0.47 breaking change)

Buthrakaur:
see https://github.com/facebook/react-native/releases/tag/v0.47.0 for explanation

### DIFF
--- a/android/src/main/java/com/rnfs/RNFSPackage.java
+++ b/android/src/main/java/com/rnfs/RNFSPackage.java
@@ -18,11 +18,6 @@ public class RNFSPackage implements ReactPackage {
   }
   
   @Override
-  public List<Class<? extends JavaScriptModule>> createJSModules() {
-    return Collections.emptyList();
-  }
-
-  @Override
   public List<ViewManager> createViewManagers(ReactApplicationContext reactContext) {
     return Arrays.<ViewManager>asList();
   }


itinance:
Thx... It is very confusing. First in. Then out.
It is breaking change up to 0.47.0 while  createJSModules will be required for 0.45 and 0.46?
Can someone confirm? @Buthrakaur @maxkomarychev @joshuapinter (according to https://github.com/itinance/react-native-fs/pull/319)

maxkomarychev:
@itinance the breaking change was introduced in recent release [0.47.0 of `react-native`](https://github.com/facebook/react-native/releases/tag/v0.47.0).
The mentioned method is no longer exists in RN api in java. This means all libraries which still have this `@Override...` are no longer compatible with latest release of RN.

I am not entirely sure if the library will still be working with previous versions of RN without this method howerver I would say this method should be removed from the mainstream so people who already updated to 0.47 can use it normally. Those who didn't can still use previous versions or fork with backports.

Buthrakaur:
I agree with @maxkomarychev according to versioning vs RN compatibility. The 0.47 dependency should be probably strictly defined in package.json though, which I didn't make.. ?

itinance:
Thx for clarification.

I will do the following: i will create a new (long overdue) release that is valid until 0.46.
After it i will then create another new release with 0.47-breaking change.

This should be more convinient for all those that will wait with upgrading to 0.47 (which is quiet common for already existing RN-projects to wait a bit with upgrading)

maxkomarychev:
@Buthrakaur I agree, you should also update version of `react-native` in `package.json` to be greater than 0.47.0.

side question: doesn't it have to defined it as `peerDependency` as well?

itinance:
At the end I handled it differently as seen in this PR: https://github.com/andpor/react-native-sqlite-storage/pull/188/commits

I added back the method but removed the **@override**.

PR #326, opened 2017-08-04
Fixing broken mTime on Android 5.x

K-Leon:
Mtime was broken - int conversion damaged Timestamp. So i changed it to putDouble which results in appropriate Timestamps.


itinance:
Thx for pointing out! But shouldn't we rather use a long-integer ("long") instead of double? Current implementations out there are used to get integers and no floating point numbers

K-Leon:
Yes, but you can't transmit longs over the RN Bridge. There are three applyable alternatives: a) int (too small) b) string (needs parsing in JS Land) c) double

itinance:
Thank you!

PR #335, opened 2017-08-22
(iOS only) Resumable downloads and better background downloads handling

ptelad:
Hi,

We've discussed about resumable downloads in my last pull request, I have not made any changes to this feature in this one.

Now about background downloads:
In my app I'm downloading large files, so bg-d/l is crucial.
I've encountered issues with bg-d/l using RNFS and found out it's because the download tasks's completion handler is not being call, thus iOS brutally terminates the app.
So I've added a new method called `completeHandlerIOS(jobId)` that will call that job's completion handler.
Now bg-d/l are safe to use and you can easily do some stuff after the download is done and then call `completeHandlerIOS()`.
readme was updated with a tutorial on how to set this up in the AppDelegate.m file.

I've also removed the bit in the readme about needing to activate 'background fetch' since it's not true, `background fetch` is for periodically connecting to the internet to fetch some data, it's not needed for downloads.

### DIFF
--- a/Downloader.h
+++ b/Downloader.h
@@ -24,7 +24,7 @@ typedef void (^ResumableCallback)();
 
 @interface RNFSDownloader : NSObject <NSURLSessionDelegate, NSURLSessionDownloadDelegate>
 
-- (void)downloadFile:(RNFSDownloadParams*)params;
+- (NSString *)downloadFile:(RNFSDownloadParams*)params;
 - (void)stopDownload;
 - (void)resumeDownload;
 - (BOOL)isResumable;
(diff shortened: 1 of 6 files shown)


DanNi0130:
@itinance Can we please merge this?

parim:
I also need to confirm the completed event in background.
Now it doesn't work properly.
Can it be fix?

itinance:
I'll look at this on this weekend and will test it and merge if all works fine!

itinance:
Tested, merged and published. Thank you all!

parim:
Great! it works well now.
thanks! @itinance 

I have one more question.
Is it possible to trigger download function - RNFS.downloadFile(...) in background mode?
(i.e. by receiving push notification)
Already I tried, but it doesn't start downloading until app is opened.

PR #339, opened 2017-08-31
fix touch() error on android

zhantx:
Since we can't change `ctime` on Android, I will just make it 0 so this parameter is optional without rising an error on Android.


itinance:
Thank you for contributing!

Issue #108, opened 2016-06-02
Cannot read property 'readDir' of undefined


johnphillips1992:
I get the error message: "Cannot read property 'readDir' of undefined"

When I add this line: 
var RNFS = require('react-native-fs');

I am testing on the IOS simulator.

savelichalex:
Are you sure that linking RNFS.xcodeproject right?

pixelgamma:
Same error here. I've linked everything using rnpm link react-native-fs

johanneslumpe:
Is this still an issue with the new 1.5 release?

mkraft:
@johanneslumpe I'm blocked by this issue in 1.5.1.

johanneslumpe:
@mkraft are you linking the project correctly? It seems a bit weird to me, because else we'd have quite a few more issues about this.

cjdell:
@mkraft @johanneslumpe I agree, this smells of the native module not being loaded into memory. Can you triple check that you have linked the .a file with the binary in Build Phases?

matdehaast:
You need to run react-native run-android/run-ios after installing before you can just reload the javascript. Reason being the module has native components that will not be installed yet if you do not do a full build first

npomfret:
I get a similar problem in android, but only on an app that's deployed to the app store.  It crashes immediately (in android).  Works fine in dev mode

```
07-29 09:07:27.915 23705-23746/? E/AndroidRuntime: FATAL EXCEPTION: mqt_native_modules
                                                   com.facebook.react.modules.core.JavascriptException: undefined is not an object (evaluating 'i.readDir'), stack:
                                                   <unknown>@629:187
                                                   o@2:517
                                                   e@2:326
                                                   i@2:188
                                                   <unknown>@628:299
                                                   o@2:517
                                                   e@2:326
                                                   i@2:188
                                                   <unknown>@365:1403
                                                   o@2:517
                                                   e@2:326
                                                   i@2:188
                                                   <unknown>@12:101
                                                   o@2:517
                                                   e@2:256
                                                   i@2:188
                                                   global code@635:9
                                                   
                                                       at com.facebook.react.modules.core.ExceptionsManagerModule.showOrThrowError(ExceptionsManagerModule.java:94)
                                                       at com.facebook.react.modules.core.ExceptionsManagerModule.reportFatalException(ExceptionsManagerModule.java:78)
                                                       at java.lang.reflect.Method.invokeNative(Native Method)
[...]

PR #341, opened 2017-09-05
Update README.md

nomoreboredom:
Requiring the module with ES6 syntax throws an error.

As this issue is also https://github.com/itinance/react-native-fs/issues/108 about the same error message, I wanted to clarify the readme for future users

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,7 @@ public class MainApplication extends Application implements ReactApplication {
 ```javascript
 // require the module
 var RNFS = require('react-native-fs');
+//Requiring the module with ES6 syntax (import RNFS from 'react-native-fs') throws 'Cannot read property 'readDir' of undefined'
 
 // get a list of files and directories in the main bundle
 RNFS.readDir(RNFS.MainBundlePath) // On Android, use "RNFS.DocumentDirectoryPath" (MainBundlePath is not defined)


README.md
```diff
@@ -178,6 +178,7 @@ public class MainApplication extends Application implements ReactApplication {
 ```javascript
 // require the module
 var RNFS = require('react-native-fs');
+//Requiring the module with ES6 syntax (import RNFS from 'react-native-fs') throws 'Cannot read property 'readDir' of undefined'
```
itinance: i'd suggest to tell both methods to requiring/import the module. One for ES5 and one for ES6. Since most projects are based on ES6 in the meanwhile we should not show the ES5 syntax anymore so popular. What do you think?

nomoreboredom: Well in my case actually I am using ES6 in my project, and use of 'import' threw error, so had to change it to 'require'. Does that makes sense?

Btw, I noticed after your reply, I see your name (as it is catchy :)) in RN fb group a lot, and recently in ethereum developers group as well. Will add over FB if it is ok?

itinance: Funny :) If you want to continue this conversation somewhere else (like FB) and miss my answer, please tag me or leave the link here that i can find it.

skovhus:
Import works fine. 

I think this PR should be closed.

PR #360, opened 2017-10-06
Expose the iOS `discretionary` flag on `downloadFile`

jamesreggio:
This PR exposes the [`discretionary` flag](https://developer.apple.com/documentation/foundation/nsurlsessionconfiguration/1411552-discretionary?language=objc), which is [recommended](https://developer.apple.com/library/content/documentation/Performance/Conceptual/EnergyGuide-iOS/DeferNetworking.html) for use in conjunction with background downloads.


itinance:
Wow! Great. Thank you very much. Would you mind providing some hints into the README also?

PR #367, opened 2017-10-23
 added a function called getAllExternalFilesDirs to get application-specific directories on all shared/external storage devices

ngrj:
Replicated [getExternalFilesDirs](https://developer.android.com/reference/android/content/Context.html#getExternalFilesDirs(java.lang.String)) to return array of application-specific directories on all shared/external storage devices where the application can place persistent files it owns. 

This is my first pull request for any project, please let me know, if you want any changes in the PR.

### DIFF
--- a/FS.common.js
+++ b/FS.common.js
@@ -201,6 +201,10 @@ var RNFS = {
     return RNFSManager.getFSInfo();
   },
 
+  getAllExternalFilesDirs(): Promise<FSInfoResult> {
+    return RNFSManager.getAllExternalFilesDirs();
+  },
+
   unlink(filepath: string): Promise<void> {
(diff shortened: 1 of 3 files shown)


FS.common.js
```diff
@@ -201,6 +201,10 @@ var RNFS = {
     return RNFSManager.getFSInfo();
   },
 
+  getAllExternalFilesDirs(): Promise<FSInfoResult> {
```
itinance: If i understood the code of "`public void getAllExternalFilesDirs(Promise promise){`" correctly it will return a promise with an array if Strings and not an `Promise<FSInfoResult>`, right?

itinance:
Thank you very much! Just one thing to look at (see my code-comment) and then i will merge.
You can add a second commit to this PR.

PR #371, opened 2017-10-28
Implement requiresMainQueueSetup to remove warning in RN 0.49

mars-lan:
Here's the warning message removed

[warn][tid:main][RCTModuleData.mm:69] Module RNFSManager requires main queue setup since it overrides `constantsToExport` but doesn't implement `requiresMainQueueSetup. In a future release React Native will default to initializing all native modules on a background thread unless explicitly opted-out of.


itinance:
thank you!

PR #379, opened 2017-11-06
Make mtime optional on ReadDirItem, change utime to mtime on StatResult

### LINKED ISSUE #378: mtime is unavailable for readDirAssets on Android and causes a crash
Using `RNFS.readDirAssets('/')` causes error `[RangeError: Invalid Date]`.

AssetManager doesn't offer any method for determining the last modified time so this value should be made optional.

josephroque:
Android and iOS both set `mtime` for `stat` method, so this type declaration must be updated.

Making `mtime` optional on `ReadDirItem` fixes #378

### DIFF
--- a/FS.common.js
+++ b/FS.common.js
@@ -33,7 +33,7 @@ type MkdirOptions = {
 
 type ReadDirItem = {
   ctime: ?Date;    // The creation date of the file (iOS only)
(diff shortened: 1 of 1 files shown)


itinance:
Thank you very much!

itinance:
just released with 2.8.5

Issue #208, opened 2016-11-17
Set connection timeout, read timeout for downloadFile


palmsey:
It would be helpful to be able to set the timeout values. It looks like the read timeout is currently hard coded at 15 seconds. I'm trying to download a large, dynamically generated file, and it sometimes takes more than the allowed time.

johanneslumpe:
Want to send a PR for this?

drunksaint:
https://github.com/itinance/react-native-fs/pull/301

PR #381, opened 2017-11-09
Respect readTimeout for downloadFile on iOS

josephroque:
`NSURLSessionConfiguration` does not have a concept of `connectionTimeout`, but it does have `timeoutIntervalForRequest` which provides similar functionality to `HttpURLConnection.setReadTimeout()`

Might close(?) issue #208, since this functionality will be as present as it can be on iOS and Android.


itinance:
Thank you very much!

PR #382, opened 2017-11-09
Respect readTimeout and connectionTimeout for downloadFile on Android

### LINKED ISSUE #368: the params connectionTimeout and readTimeout not assigned value in RNFSManager.java
the params connectionTimeout and readTimeout not assigned value in RNFSManager.java

josephroque:
These values were asked for, but never actually set in `DownloadParams` so they are not respected when downloading a file.

Fixes #368


itinance:
THank you very much!

PR #383, opened 2017-11-15
Fix reference path to ReactNative.csproj

pechkin:
ReactNative project referenced incorrectly. Visual Studio fixes this reference automatically, once you open your project, so this issue doesn't show up. We see this issue when we build our project with MSBuild on staging server.


itinance:
Thank you very much!

PR #386, opened 2017-11-16
Avoid placing nil value into NSDictionary in downloader

jamesreggio:
Simple bug, simple fix.

The [`NSHTTPURLResponse
.allHeaderFields`](https://developer.apple.com/documentation/foundation/nshttpurlresponse/1417930-allheaderfields?language=objc) property can return `nil` if the request experiences a connection error. If we try to place this `nil` value into the `NSDictionary` literal, we get an exception and a crash.

The fix is simply to insert the `NSNull` singleton instead, which will surface the `null` value to the JavaScript.


itinance:
Thank you very much!

PR #413, opened 2018-01-04
catch error in downloadFile

wkrause13:
calling downloadFile while offline results in an uncaught error. This pull request catches that error and returns a promise rejection.


adamvert:
This totally fixed a repeating crash that was driving me crazy - thank you!

PR #416, opened 2018-01-12
Add UWP tests to NPM ignore

rozele:
The `react-native link` command will unintentionally link the `RNFS.Tests` project instead of the `RNFS` project. This adds `RNFS.Tests` to `.npmignore` so this is no longer a problem.


itinance:
Thank you!

PR #434, opened 2018-02-06
constants.put(RNFSTemporaryDirectoryPath,null);

wsong910:
JS will get path like "null/{dir}" . Change suggestion.


itinance:
Good catch! This way `RNFSTemporaryDirectoryPath` would point to the caching directory path of `RNFSCacheDirectoryPath`. Thank you!

itinance:
Published.

PR #445, opened 2018-02-17
Fix Readme

reimertz:
..  to not give the impressions cocoapods doesn't work.


itinance:
Thx for pointing out!

PR #447, opened 2018-02-20
Fix framework target

ethanx94:
Having issues building with the newest react-native-windows.
For some reason having a target framework of "uap10.0" is invalid. I instead pointed it at 10240 and my build issues were solved. I think there was some breaking change happening either within react-native's ecosystem or UWP apps since the last sdk release.

### DIFF
--- a/windows/RNFS/project.json
+++ b/windows/RNFS/project.json
@@ -3,7 +3,7 @@
     "Microsoft.NETCore.UniversalWindowsPlatform": "5.2.2"
   },
   "frameworks": {
-    "uap10.0": {}
+    "uap10.0.10240": {}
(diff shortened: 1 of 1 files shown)


ethanx94:
See a similar issue I had with rnsqlite-2 https://github.com/craftzdog/react-native-sqlite-2/pull/24

jpshelley:
@ethanx94 does it make sense to merge this https://github.com/itinance/react-native-fs/pull/448 PR instead of this one now?

ethanx94:
Yes that seems to do the trick in a much more elegant way. I'd say take that one over this one. Thanks @rozele!

PR #460, opened 2018-03-15
Wrong return type in getAllExternalFilesDirs signature

lmvco:
Instead of string, it should be array of strings


itinance:
good catch! thx!

PR #466, opened 2018-03-27
Adding a new version of getFsInfo method 

lmvco:
With this new method, it's possible to get information about any mounted volume (just pass a file from the desired volume). It's useful in Android in order to get info about an external volume (e.g. sdcard)


lmvco:
This method is just was just implemented for Android

Issue #464, opened 2018-03-22
writeFile faild with no error


wd:
I'm using redux-persist-fs-storage which use react-native-fs as it's persist storage, I have tried to save some settings for my app but failed. After debug, I found the problem was caused by this block `https://github.com/itinance/react-native-fs/blob/master/android/src/main/java/com/rnfs/RNFSManager.java#L128-L130` of code, it will be success without error, but the file not exists acutely.

My environments:
react-native: 0.52.3
react-native-fs: 2.9.11
Android: 7.1.1
Pad model: Lenovo TB-X704N

Burgomehl:
Same for me with react-native 0.52.2

wd:
I have issued a pull request about this problem.

Here is a quick fix for those who didn't want to wait the merge. When you use `redux-persist`, you can use `keyPrefix` for your persist config  to avoid this problem. The default `keyPrefix` is `persist:`, avoid use any `urlencode/urldecode` chars will not cause this problem.

```
const userPersistConfig = {
    key: 'User',
    storage: storage,
    keyPrefix: ''
}
```

itinance:
Please read my comment and suggestions towards this issue in the specific PR from @wd https://github.com/itinance/react-native-fs/pull/468#issuecomment-380806481

neverPick:
Maybe you need to config provider when running on android 7.0+. And U need to hack `getFileUri` of `RNFSManager,`  ` File saveFile=new File(filepath);
    Uri uri = FileProvider.getUriForFile(this.reactContext,
            this.reactContext.getApplicationContext().getPackageName() + ".provider",saveFile);`

PR #468, opened 2018-03-30
Add urldecode when check if the file exists

wd:
As #464 mentioned,  If the file save with a name like `files/persist/persist%3AUser`, the file will be actually saved as `files/persist/persist:User` on phone storage. So this should be do the same to the `exists` method.

I haven't test much on this patch(only test with `redux-persist-fs-storage`, it works), hope it will work on all conditions.

### DIFF
--- a/android/src/main/java/com/rnfs/RNFSManager.java
+++ b/android/src/main/java/com/rnfs/RNFSManager.java
@@ -178,7 +178,7 @@ public class RNFSManager extends ReactContextBaseJavaModule {
   @ReactMethod
   public void exists(String filepath, Promise promise) {
     try {
-      File file = new File(filepath);
+      File file = new File(URLDecoder.decode(filepath, "UTF-8"));
       promise.resolve(file.exists());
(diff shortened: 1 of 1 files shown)


itinance:
Thank you for this PR. 

I know what you try to solve and this would be a good fix for the colon-issue, but I am not sure if we would break something here, because in all other methods like "readFile" and so on nowhere the `decode`-function is used yet:

`File file = new File(directory);`

against to 

`File file = new File(URLDecoder.decode(filepath, "UTF-8"));`

Shouldn't we use the decode-function everywhere to be consistent?

But if we would do so, there could be compatibility issues towards those files that has been created in older versions of react-native-fs.
I'd like to suggest to use the decode-function on all lines where a File-object will be created with a path-argument, but also we should provide a feature-flag, that will allow to switch to the "old"-mode. This feature flag would be placed best in the context of the RNFSManager instead of adding a separate argument to all those specific methods-

What do you think?

PR #480, opened 2018-04-19
Android: Resolve filepath for content URI

andtos90:
File sharing with the content URI will not use the original filename, I've just updated the `stat` API to return the original file path (the file pointed by the content uri).

Example:
screen shot 2018-04-19 at 19 09 27

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,8 @@ DerivedData
 # Pods - for those of you who use CocoaPods
 Pods
 update-test.sh
+.vscode/
+android/.gradle/*
+android/gradle/*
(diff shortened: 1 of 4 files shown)


andtos90:
Done, it's a brief introduction but I hope that explains the correct use case

elribonazo:
I had a problem also with paths between the original file path and the one react-native is able to load.
How i finally solved it was adding file:// to the originalFilePath, i'm not sure of your requirements but I can work with the images perfectly now

elribonazo:
@itinance Can we please plan the merge / pull request please?

Issue #499, opened 2018-06-05
uploadFiles leaks memory


ainth:
In `RNFSManager.m` in `uploadFiles` the new upload job is added to the `self.uploaders` dictionary. Jobs are never cleaned up and when uploading a large number of large files this will eventually crash the device (as in our case) as it seems the body of the request is retained.

I've confirmed this in xcode's instruments where I see `NSMutableData` objects being retained that are created from uploadFiles that are persisted and never get cleaned up.

itinance:
Fixed by https://github.com/itinance/react-native-fs/pull/502

PR #502, opened 2018-06-08
fix(upload): prevent memory leaks from retained jobs

Ignigena:
Ran into the same issue as #499 particularly since our application uploads a number of files and is rarely re-launched. This PR simply removes the job from the `NSMutableDictionary` as soon as it either succeeds or fails.


itinance:
Good catch, many thanks!

PR #503, opened 2018-06-11
make sure to return the original file uri if content:// scheme is not used

baabgai:
If the stat function was called on Android using a path like file://... the getOriginalFilePath function would just return an empty string which resulted in a 'File does not exist exception'.
So if the original path is given as a file:// scheme just leave it be.


superandrew213:
Getting same issue.  `stat` currently only supports `content://...` paths on Android.

itinance:
Thank you!

Issue #509, opened 2018-06-24
`IndexOutOfBoundsException` when uploading files on Android


aichamorro:
# Overview
I just integrated `react-native-fs` into my React Native application and, while iOS file upload works well, I was having problems while uploading files in Android (using the same API). Specially when the file is smaller than the allocated buffer.

# The problem
According to the documentation of [FileInputStream](https://developer.android.com/reference/java/io/FileInputStream.html#read()), an `IndexOutOfBoundsException` is thrown in the following cases:
  1. `off` is negative
  2. `len` is negative
  3. `len` is greater than `b.length - off`

The first and the second cases don't seem to be likely to happen in [Uploader.upload](https://github.com/itinance/react-native-fs/blob/master/android/src/main/java/com/rnfs/Uploader.java#L48), but I found the third case may happen since the code is not making sure the `FileInputStream` is not read beyond its available data. Take this for instance (taken from [here](https://github.com/itinance/react-native-fs/blob/master/android/src/main/java/com/rnfs/Uploader.java#L110)):
```
bufferAvailable = 4096;
bufferSize = Math.min(bufferAvailable, maxBufferSize);
byteRead = fileInputStream.read(b, 0, bufferSize);
```
If the file is smaller than `bufferSize`, means that `bufferSize > b.length - offset` and that causes an `IndexOutOfBoundsException` to be thrown. The same happens inside the (loop)[https://github.com/itinance/react-native-fs/blob/master/android/src/main/java/com/rnfs/Uploader.java#L115-L126]:
```
while (byteRead > 0) {
  if (mAbort.get()) throw new Exception("Upload has been aborted");
  
  request.write(b, 0, bufferSize);
  byteRead = fileInputStream.read(b, 0, bufferSize);
  if (byteRead == -1) {
    mParams.onUploadProgress.onUploadProgress(fileCount, totalSize, Readed);
  } else {
    Readed += byteRead;
    mParams.onUploadProgress.onUploadProgress(fileCount, totalSize, Readed);
  }
}
```
Whenever a read operation is performed, there should be a check to ensure it's not reading beyond the limits of the `InputStream`.

Environment: 
React Native: 0.55.4
Android SDK: 26

PR #510, opened 2018-06-24
Fixes an `IndexOutOfBounds` while uploading files in Android

aichamorro:
# Overview
Fixes https://github.com/itinance/react-native-fs/issues/509.

# Description
According to the documentation of FileInputStream, an IndexOutOfBoundsException is thrown in the following cases:
  1. off is negative
  2. len is negative
  3. len is greater than `b.length - off`

The third case was the one that was throwing an exception when reading from the `FileInputStream`. More information in the [issue](https://github.com/itinance/react-native-fs/issues/509)


hank121314:
Sorry for that I didn't test that much before I make last uploadFiles PR. 
Big thanks for this PR.

aichamorro:
Thank you for investing your time in this. I'm glad I can help making it a bit better.

itinance:
Thank you guys very much!

PR #516, opened 2018-07-06
[iOS] Ensure _bytesWritten is correct in download completion callback

jamesreggio:
As the comment says, there's no guarantee that the `URLSession:downloadTask:didWriteData:` delegate callback will ever fire, and thus it's possible for a download to finish without `_bytesWritten` ever being set.

If the download finished successfully (as it did if we reach this callback), there's no reason not to just read the size from the filesystem.


itinance:
Thank you vey much!

Issue #517, opened 2018-07-09
No resource found that matches the given name: attr 'android:keyboardNavigationCluster'.


banli17:
a new project rn0.56，just install react-native-fs。when run react-native run-android ,error follow

```
node_modules/react-native-fs/android/build/intermediates/res/merged/release/values-v26/values-v26.xml:15:21-54: AAPT: No resource found that matches the given name: attr 'android:keyboardNavigationCluster'.
```

banli17:
i found it ，sdk version  issue。

RodolfoSilva:
@banli17 How did you fix it? Share please, I'm having the same problem.

RodolfoSilva:
Solution: https://github.com/itinance/react-native-fs/issues/507#issuecomment-403198880 and #519

The-Code-Monkey:
Update your version as the lastest pr that I made was merged and it fixed it

PR #519, opened 2018-07-10
updated compilesdkversion and buildtoolsversion

The-Code-Monkey:
updated the build.gradle to 26 and 26.0.3

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -11,8 +11,8 @@ buildscript {
 apply plugin: 'com.android.library'
 
 android {
-    compileSdkVersion 25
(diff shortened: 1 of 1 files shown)


RodolfoSilva:
@itinance I think we need to generate a new version with this PR, because any project with react-native0.56.0 will be broken when trying to use this project.

This PR will solve #517

vieiralucas:
How long till this makes to npm?

RodolfoSilva:
@itinance :point_up:

ezailWang:
@itinance ☝️☝️

PR #535, opened 2018-08-07
Make this work with RN56

rmevans9:
This makes it so users can configure the versions via the new properties RN is adding to the build.gralde


itinance:
Wow! Good catch. Thank you!

PR #541, opened 2018-08-17
IOS: File Protection Support

gensc004:
Adding support for specifying IOS File Protections when creating/moving/copying files, and making directories.

### DIFF
--- a/FS.common.js
+++ b/FS.common.js
@@ -29,6 +29,11 @@ var normalizeFilePath = (path: string) => (path.startsWith('file://') ? path.sli
 
 type MkdirOptions = {
   NSURLIsExcludedFromBackupKey?: boolean; // iOS only
+  NSFileProtectionKey?: string; // IOS only
+};
+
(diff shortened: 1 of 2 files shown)


itinance:
Pretty cool! Thank you very much!

AlexanderZaytsev:
@gensc004 could you please provide an example on how to use it? Should it be something like this?
```javascript
RNFS.writeFile(path, data, { NSFileProtectionKey: 'NSFileProtectionComplete'})
```

gensc004:
@AlexanderZaytsev Unfortunately its been a while since I made this PR, and I don't have an example of implementing it my project anymore. Based on what I remember, and looking through the code again, it does look like that example you gave should work! Did you get a chance to test it out?

Fannolo:
@AlexanderZaytsev from what I saw in the code, correct me if I'm wrong, writeFile does not support NSFileProtectionKey, also DownloadFile seems to have the same problem.

Issue #79, opened 2016-04-19
Android: Media Scanner


paulondc:
Hi,

I noticed that when new images (or any file) are created under a location aka PicturesDirectoryPath they don't become visible immediately under the gallery app. The reason for that seems it requires an extra step to tell the android's mediascanner about the new files. By not doing that means the new files are going to get "visible" for users only when mediascanner starts which happens during the boot.

Following bellow a stackoverflow about this issue:
http://stackoverflow.com/questions/9414955/trigger-mediascanner-on-specific-path-folder-how-to

Cheers

johanneslumpe:
@paulondc Is there anything we can do here?

paulondc:
Hi @johanneslumpe,

I believe by having a way to trigger the mediascanner through native-fs would be beneficial for the library, I can see this feature being useful for any application that create files under shared places which are meant to be shared across different apps (like pictures), therefore this feature would be in line with the library, since this just makes the files generated by them available in [android's mediastore](http://developer.android.com/intl/es/reference/android/provider/MediaStore.html):

I was thinking something like:
```java
  @ReactMethod
  public void updateMediaStore(String path, Callback callback) {
    ...
  }
```

What do you think about it ?

johanneslumpe:
@paulondc I'm not sure that this really belongs into the library. I'd kind of like to open this one up for discussion and see if other people think it's useful before we make a decision on this.

pewh:
+1 this is very useful for me. I cannot see downloaded image at Gallery unless restart the device :sob:

havir:
+1

hellomaya:
When I used Android File Transfer to check the files on Android, and the folder wasn't there after mkdirs, due to this reason. It's necessary to show the folder to user, somehow, they can download their assets directly there, or the app will be complex to have function to be the same for them.

If the folder or file changes can be shown after it created, or even after the device reboot, will be good, looks like it can't for now.

Anyway, to build a native module by adding MTP scan API out of this lib is also possible, but if this lib can provide this as separate API sets, will be very helpful.

I believe this problem also happened in deleting a file, for example Image, then we can't see this change in system gallery or else.

phillmill:
+1

lefoy:
[...]

PR #544, opened 2018-08-20
Add scanFile method (Android only)

lefoy:
Added scanFile method for Android only. Should solve issue #79


itinance:
Many thanks!

PR #557, opened 2018-09-27
copyFile: fix missing parameter on Android.

cknitt:
Currently, my Android app crashes when I invoke copyFile, because the `options` parameter is missing.

JS:
```js
copyFile(filepath: string, destPath: string, options: FileOptions = {}): Promise<void>
```

iOS:
```objc
RCT_EXPORT_METHOD(copyFile:(NSString *)filepath
                  destPath:(NSString *)destPath
                  options:(NSDictionary *)options
                  resolver:(RCTPromiseResolveBlock)resolve
                  rejecter:(RCTPromiseRejectBlock)reject)
```

but Android:
```java
public void copyFile(String filepath, String destPath, Promise promise)
```

### DIFF
--- a/android/src/main/java/com/rnfs/RNFSManager.java
+++ b/android/src/main/java/com/rnfs/RNFSManager.java
@@ -334,7 +334,7 @@ public class RNFSManager extends ReactContextBaseJavaModule {
   }
 
   @ReactMethod
-  public void copyFile(String filepath, String destPath, Promise promise) {
+  public void copyFile(String filepath, String destPath, ReadableMap options, Promise promise) {
     try {
       copyFile(filepath, destPath);
 


famousfilm:
@cknitt Would you mind adding support for moveFile as well? It is broken for the same reason because of this PR: https://github.com/itinance/react-native-fs/pull/541/files ... looking at that, writeFile might be broken, too. 
They seem like the same issue (so should be fixed by the same PR), but I'm happy to do it if you don't think so. Just ping me.

famousfilm:
...nm it's already merged in @cknitt

itinance:
I will fix moveFile as well, thanks for your hints!

itinance:
#558

itinance:
released. thanx guys!

PR #564, opened 2018-10-07
Replace deprecated 'compile' gradle configuration with 'implementation

mars-lan:
See http://d.android.com/r/tools/update-dependency-configurations.html for more details

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -30,5 +30,5 @@ android {
 }
 
 dependencies {
-    compile 'com.facebook.react:react-native:+'
+    implementation 'com.facebook.react:react-native:+'
 }


itinance:
thx!

skb1129:
@itinance This is a breaking change, it works with Gradle version 3+ only and Gradle 3+ is available in react native version 0.57.0+

itinance:
@skb1129 any advice how to make it backwards compatible?

itinance:
One option could be that older RN versions should be fixed with v2.11.17:
 https://github.com/itinance/react-native-fs/tree/v2.11.17

skb1129:
@itinance Yeah older RN versions can use v2.11.17
You should release v2.12.0 for these changes.

skb1129:
And also mention this as a breaking change in the release notes

Issue #566, opened 2018-10-12
[Android] Upload complete, but progress callback not sync


jonathangreco:
## Context

I upload on a Symfony 4 backend some videos, from cameraRoll.

## Issue
My server received all the videos, but the progress is still at 18% and keep progressing.
The callback on upload progress show bytesWritten , I suspect it to be wrong.

## Problem
This is an issue because the progress is not sync with what's really sent to the server. Bytes sent are wrong.

## Code
```

  componentDidMount() {
    let filetype = 'video/mov';
    let extension = ".mov";
    let uploadFiles = [];


    if (Platform.OS === "android") {
      filetype = 'video/mp4';
      extension = '.mp4';
    }

    const {localTemplate} = this.props;

    if (localTemplate.status && localTemplate.status !== "pending") {

      let statsFiles = localTemplate.scenes.map(async (scene) => {
        return RNFS.stat(scene.film).then(result => {
          return {...result, scene};
        });
      });

      uploadFiles = Promise.all(statsFiles).then((files) => {
        return files.map((file) => {
          let filePath = file.path.replace('file:///', '');
          if (Platform.OS === "android") {
            filePath = file.originalFilepath;
          }

          return Object.assign({
            name: file.scene.name,
            filename: file.scene.name + extension,
            filetype: filetype,
            filepath: filePath
          })
        })
      });

      uploadFiles.then(async (files) => {
        let token = await AsyncStorage.getItem("token");
        RNFS.uploadFiles({
          toUrl: GLOBALS.BASE_URL + '/api/templates/submit/rush',
          files: files,
          method: 'POST',
          fields: {
            'template': "" + localTemplate.id
          },
          headers: {
            'Accept': 'application/json',
            Authorization: 'Bearer ' + token
          },
          begin: this.uploadBegin,
          progress: this.uploadProgress
        }).promise.then((response) => {
          if (response.statusCode !== 200) {
            Toast.show({
              text: response.message,
              duration: 5000,
              type: 'danger'
            })
          } else {
            let body = JSON.parse(response.body);
            this.saveTemplate(body.index)
          }
        })
        .catch((err) => {
          this.setState({
            uploadBeginText: I18n.t('error_during_process'),
            uploadFinished: true,
            uploadFinishedText: I18n.t('retry_or_contact_us')
          });
          Toast.show({
            text: err.message,
            duration: 5000,
            type: 'danger'
          })
        });
      });
    } else {
      this.setState({
[...]

PR #571, opened 2018-10-26
Fix issue #566 android progress callback not sync and handle uppercase file extension mimetype

hank121314:
### Problem
1. This is an issue because the progress is not synced with what's really sent to the server. Bytes sent are wrong.
2. If the file extension is uppercase mimetype will return null.
### Reason
1. Two situations might cause this issue:

  - 1. HttpURLConnection.setFixedLengthStreamingMode should be called,and it's parameter should 
     be request body length.
  - 2. It might be react-native bugs. When I was sending too much event to sendEvent, it will have 
     some delay.
2. Android package MimeUtils cannot get mimetype when the file extension is uppercase.
### Resolve
1.
   - 1. Call setFixedLengthStreamingMode and calculate correct content-length.
   - 2. Send less event to sendEvent.
2. Add lowerCase to file extension when getMimeType

### DIFF
--- a/android/src/main/java/com/rnfs/RNFSManager.java
+++ b/android/src/main/java/com/rnfs/RNFSManager.java
@@ -734,11 +734,10 @@ public class RNFSManager extends ReactContextBaseJavaModule {
       };
 
       params.onUploadProgress = new UploadParams.onUploadProgress() {
-        public void onUploadProgress(int fileCount,int totalBytesExpectedToSend,int totalBytesSent) {
+        public void onUploadProgress(int totalBytesExpectedToSend,int totalBytesSent) {
           WritableMap data = Arguments.createMap();
 
           data.putInt("jobId", jobId);
-          data.putInt("FileID",fileCount);
           data.putInt("totalBytesExpectedToSend", totalBytesExpectedToSend);
           data.putInt("totalBytesSent", totalBytesSent);
 
--- a/android/src/main/java/com/rnfs/Uploader.java
+++ b/android/src/main/java/com/rnfs/Uploader.java
@@ -16,7 +16,6 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.InputStreamReader;
 import java.net.HttpURLConnection;
-import java.nio.charset.Charset;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicBoolean;
--- a/android/src/main/java/com/rnfs/Uploader.java
+++ b/android/src/main/java/com/rnfs/Uploader.java
@@ -93,7 +93,7 @@ public class Uploader extends AsyncTask<UploadParams, int[], UploadResult> {
                 File file = new File(map.getString("filepath"));
                 String fileHeaderType = twoHyphens + boundary + crlf +
                         "Content-Disposition: form-data; name=\"" + name + "\"; filename=\"" + filename + "\"" + crlf +
-                        "Content-Type: " + filetype + crlf +"Content-Transfer-Encoding: binary" + crlf;
+                        "Content-Type: " + filetype + crlf;
                 long fileLength = file.length();
                 totalFileLength += fileLength ;
                 if(params.files.toArray().length - 1 == fileCount){
--- a/android/src/main/java/com/rnfs/Uploader.java
+++ b/android/src/main/java/com/rnfs/Uploader.java
@@ -58,13 +58,10 @@ public class Uploader extends AsyncTask<UploadParams, int[], UploadResult> {
         long totalFileLength = 0;
         BufferedInputStream responseStream = null;
         BufferedReader responseStreamReader = null;
-        int maxBufferSize = 1 * 1024 * 1024;
         String name, filename, filetype;
         try {
             connection = (HttpURLConnection) params.src.openConnection();
-            connection.setUseCaches(false);
             connection.setDoOutput(true);
-            connection.setDoInput(true);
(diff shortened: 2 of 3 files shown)


jonathangreco:
@itinance Hi, can we merge it during the next week ? Is it possible in a delay like this ?

Thanks in advance for this :)

android/src/main/java/com/rnfs/RNFSManager.java
```diff
@@ -734,11 +734,10 @@ public void onUploadBegin() {
       };
 
       params.onUploadProgress = new UploadParams.onUploadProgress() {
-        public void onUploadProgress(int fileCount,int totalBytesExpectedToSend,int totalBytesSent) {
```
itinance: Is it a good idea to remove the FileID? What if i would to assign the upload progress event to a particular File upload running multiple of them?

android/src/main/java/com/rnfs/Uploader.java
```diff
@@ -117,13 +116,13 @@ private void upload(UploadParams params, UploadResult result) throws Exception {
             for (ReadableMap map : params.files) {
                 request.writeBytes(fileHeader[fileCount]);
                 File file = new File(map.getString("filepath"));
-                FileInputStream fis = new FileInputStream(file);
                 int fileLength = (int) file.length();
                 int bytes_read = 0;
                 int bytesReadTotal = 0;
-                int buffer_size = fileLength / 100;
+                int buffer_size =(int) Math.ceil(fileLength / 100.f);
```
itinance: I think we should use a fixed buffer size depending on the available memory size of the device instead of 1 percent of the file length. If the user wants to upload a very big file, the app could crash because of "Out of memory" exception.
Since this file length based approach was already present before this PR, it has not to be done here. Just an idea for further discussions and for next steps :) 
What do you think?

android/src/main/java/com/rnfs/RNFSManager.java
itinance: Also, the jobID-attribute is documented in README and part of the iOS-implementation also. Therefor we should keep it.

itinance:
@jonathangreco i am already on it :) 

@hank121314 
Just added some hints. In particular, to remove the fileID/JobId  from onProgress-Event is against our documentation and would differ here from iOS-implementation. We should keep it to not break any existing software.

android/src/main/java/com/rnfs/RNFSManager.java
hank121314: Because IOS uploadFiles don't have FileID.
In this version, totalBytesSent and totalBytesExpectedToSend are depended on the file which is uploading now.
But IOS totalBytesSent and totalBytesExpectedToSend is depended on whole request length.
I think we should reach a consensus about the progressCallBack 😄

jonathangreco:
@itinance Thanks i saw I've read all of your review messages. Waiting :)

Thanks for your time, really.

android/src/main/java/com/rnfs/RNFSManager.java
itinance: Ah yes, probably a misconception between jobID and fileID.

on iOS (and in Readme), the jobID is mandatory:
https://github.com/itinance/react-native-fs/blob/master/RNFSManager.m#L601

itinance: As long as we keep this also by for Android i am fine with the changes.

hank121314: Yeah!
I still keep jobID in progressCallBack.
[https://github.com/hank121314/react-native-fs/blob/master/android/src/main/java/com/rnfs/RNFSManager.java#740](url)

android/src/main/java/com/rnfs/Uploader.java
hank121314: That's a big problem. 😱 
I will find a solution!
Thanks for your reminder!

itinance: What do you think, shall we solve this buffer size issue with another PR in favour of a fast merge of this one, @hank121314 ?

hank121314: Could you give me one day to fix this issue?
I think I can resolve this quickly!

itinance: Sure. Looking forward

hank121314: Hi! I use Runtime.freememory() to compare with fileLength /  100.
I think this might be solved OOM issue.

jonathangreco:
@hank121314 @itinance don't forget me guys, please :)

itinance:
I'm ready to create a new release. Would you like to join testing current master with merged PR, @jonathangreco ?

jonathangreco:
@itinance Yes of course, I can test on my own app, will this be enougth ?

itinance:
Sure :) Which one is it? i'm testing the same here with https://zizzle.io :)

jonathangreco:
not realeased yet but it's https://www.dollycast.io

itinance:
v2.12.1 is out now

Thank you very much @hank121314 and @jonathangreco

jonathangreco:
@itinance Hi, thank you :) and good job for this release, thanks for your time.

@hank121314 Really thank for your fix. Did a great Job.

VishalTennyson:
Still the progress callback is not synchronized with the server.The server finishes uploading in 50% of the progress itself but it still runs twice the time.

PR #579, opened 2018-11-01
Update README.md

riccardonuzz:
I added a note just to make more clear that when using some directories it's necessary to ask for permissions to the operating system (tested on Android)

### DIFF
--- a/README.md
+++ b/README.md
@@ -362,6 +362,8 @@ The following constants are available on the `RNFS` export:
 - `ExternalDirectoryPath` (`String`) The absolute path to the external files, shared directory (android only)
 - `ExternalStorageDirectoryPath` (`String`) The absolute path to the external storage, shared directory (android only)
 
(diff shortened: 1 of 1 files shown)


itinance:
Thank you!

riccardonuzz:
I just got a question: documentation says that current usage is limited to the DocumentDirectory inside app but asking for permissions seems to make this library work in every other directory (i read my phone's storage). Is that a documentation error? Didn't try to write a file anyway. Here is the snippet of code that i used:

```
import { PermissionsAndroid } from 'react-native';

PermissionsAndroid.request(PermissionsAndroid.PERMISSIONS.WRITE_EXTERNAL_STORAGE, {
    title: 'Download',
    message: 'Allow me to download a file.'
}).then(granted => {
    if (granted === PermissionsAndroid.RESULTS.GRANTED) {
        console.log('You can use filesystem');
        readDir(ExternalStorageDirectoryPath).then(result => {
            console.log(result);
        });
    } else {
        console.log('Permission denied');
    }
})
.catch(err => console.warn(err));
```

AbdelkaderOumrani:
it is not working on android 10 even after asking for permissions

PR #581, opened 2018-11-12
Windows fixes for RN 0.47.*

juhasuni:
Batch of changes that fix compatibility with the latest `react-native-fs` API, `react-native-windows` API and dev environment.


itinance:
thx man!

PR #587, opened 2018-11-19
Fix native Windows implementation writeFile

aikewoody:
Use the correct parameters for the native Windows implementation of the 'writeFile' function.


itinance:
thx!

Issue #542, opened 2018-08-19
RNFS.copyFile() block/freeze the UI


ggunti:
RNFS.copyFile() freezes the UI.

My code:
`  copyFiles(files) {
    const path = `${RNFS.DocumentDirectoryPath}/${this.root}`;
    // Make the directory
    RNFS.mkdir(path, { NSURLIsExcludedFromBackupKey: true });

    const promises = [];
    const distinctFiles = [];
    for (const file of files) {
      promises.push(RNFS.exists(path + file.fileName).then(e => {
        if (!e) {
          console.log(path + file.fileName);
          distinctFiles.push(file);
          return RNFS.copyFile(file.path, path + file.fileName);
        }
      }));
    }
    Promise.all(promises).then(() => {
      this.props.addFiles(distinctFiles);
      this.props.addFilesFinished();
    }).catch(() => {
      this.props.addFilesFinished();
    });
  }`

The UI is blocked (I can press anything, it not work) until all the files are copied (using RNFS.copyFile(..., ...)).

Expectation: to copy all the files asynchronously and the UI to be not blocked/frozen while copying the files.

PR #597, opened 2018-12-14
Perform copyFile in background thread to prevent UI blocking

EdTorbett:
This PR fixes issue #542.


EdTorbett:
It's worth noting that the observed issue either only affects Android and not iOS, or file copies on iOS are so much faster that the issue is not apparent. This PR has been tested with files around 200MB in size and reduces UI blocking from ~10 seconds to zero.

itinance:
Thx a lot!

PR #599, opened 2018-12-17
Update RNFSManager.m

robyweb:
This adds support for the retrieval of iCloud stored videos. The only addition was allowing network access to retrieve the video asset.


itinance:
Thx a lot!

PR #622, opened 2019-01-25
[WIP] [UWP] Add encryption capabilities

leoek:
**This PR provides file encryption for UWP apps.**

It should be easy to adapt the encryption capabillity from these functions to the other ones (see Todo section below) as well. (Essentially it's just a check whether encryption is required or not and using the crypto stream instead of the filestream)

Unfortunately there are two minor changes in [FS.common.js](https://github.com/leoek/react-native-fs/blob/2d9b481782bd508876674d305ba83ab029292197/FS.common.js) which affect the functionality of android and ios. We need an extra options object for readFile and downloadFile. Android and ios can ignore the property until the native code for them is ready. For writeFile, this is already present and will only have an additional property.

**UWP - Todos:**
- [x] writeFile
- [ ] appendFile
- [ ] write
- [x] readFile
- [ ] read
- [x] downloadFile

**Android - Todos (Compatibillity only):** 
- [ ] readFile
- [ ] downloadFile

**Ios - Todos (Compatibillity only):**
- [ ] readFile
- [ ] downloadFile

This PR handles the implementation of the encryption for UWP only. The implementation for android and ios should get their own PRs.


enrico-hic:
Is it possible for you to align the PR to the new master? It would be very useful to use it with the newer version of the library

PR #648, opened 2019-04-02
Add null check for getExternalFilesDirs paths

### LINKED ISSUE #478: NullPointerException on getAllExternalFilesDirs
Hi,
My app analytics are showing a large number of crashes when calling getAllExternalFilesDirs. This isn't something I've managed to yet reproduce locally, but it seems to be hitting a significant percentage of my users. The message is:
```
com.rnfs.RNFSManager.getAllExternalFilesDirs
RNFSManager.java - line 714
java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String java.io.File.getAbsolutePath()' on a null object reference
```
The code in question is:
```jsx
export async function getSdCardDirectory() {
    if (Platform.OS === "android") {
        const paths = (await RNFS.getAllExternalFilesDirs()).filter(
            (path) => !/emulated/.test(path)
        );
        if (paths.length > 0) {
            return paths[0];
        }
    }

    return RNFS.ExternalDirectoryPath || RNFS.DocumentDirectoryPath;
}
```
For context, all of the devices that run this are identical, and all have an SD card.

ascorbic:
The array returned by `getExternalFilesDirs` [may include null elements](https://developer.android.com/reference/android/content/Context.html#getExternalFilesDirs(java.lang.String)) if the device isn't currently available. If this happens currently, it causes a NullPointerException. This PR adds a null check.

Closes #478

### DIFF
--- a/android/src/main/java/com/rnfs/RNFSManager.java
+++ b/android/src/main/java/com/rnfs/RNFSManager.java
@@ -898,7 +898,9 @@ public class RNFSManager extends ReactContextBaseJavaModule {
     File[] allExternalFilesDirs = this.getReactApplicationContext().getExternalFilesDirs(null);
     WritableArray fs = Arguments.createArray();
     for (File f : allExternalFilesDirs) {
-      fs.pushString(f.getAbsolutePath());
+      if (f != null) {
+        fs.pushString(f.getAbsolutePath());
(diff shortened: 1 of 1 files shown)


itinance:
Thanks for this!

PR #654, opened 2019-04-10
bugfix(android): clearly error message

zjkhiyori:
When i came across "Permission denied" error through RNFSManager.write(), 
it throw "no such file or directory", that made me confusion, 
change to this to throw the true error :)


itinance:
thx!

PR #661, opened 2019-04-23
#660 add a custom error for cancelled downloadFile Requests

### LINKED ISSUE #660: [UWP] downloadFile Promise should reject with a proper error if cancelled.
Currently there is just a [default Exception Handler](https://github.com/itinance/react-native-fs/blob/57b6131aaad5b7721857b9bf24dc86d8bcaf6eda/windows/RNFS/RNFSManager.cs#L462) which handles all exceptions which can occur. However cancellation is an expected error and should therefore have its own customized error.

leoek:
This fixes #660 

There shouldn't be any compatibility issues with android and ios. This just makes the error for cancelled windows requests more specific.

### DIFF
--- a/windows/RNFS/RNFS.csproj
+++ b/windows/RNFS/RNFS.csproj
@@ -126,6 +126,7 @@
   </PropertyGroup>
   <ItemGroup>
     <Compile Include="Properties\AssemblyInfo.cs" />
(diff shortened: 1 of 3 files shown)


itinance:
Thanks for this!

Issue #461, opened 2018-03-15
sdcard info in getfsinfo response


lmvco:
The [getFSInfo](https://github.com/itinance/react-native-fs#getfsinfo-promisefsinforesult) method only returns information about device internal memory.

Will this FSInfoResult object include information about sdcard (for Android devices with sdcard mounted)?

yususs:
Hi. Did you find any  solution?

naxel:
created pull-request https://github.com/itinance/react-native-fs/pull/669

PR #669, opened 2019-05-01
Changed "size" from Int to Double

naxel:
Changed stat()'s and readDir()'s "size" from Int to Double
For big files


naxel:
please accept this :)

naxel:
Added to FSInfoResult object information about sdcard (for Android devices with sdcard mounted)?
#461
```js
const fsInfo = await RNFS.getFSInfo();
// Android
const { totalSpace, freeSpace,  totalSpaceEx, freeSpaceEx } = fsInfo;
```

PR #670, opened 2019-05-08
Fix error with copy file with spaces in title

YozhikM:


jsamr:
@YozhikM RNFS is behaving appropriately. You should URL-decode the file name. Perhaps you run into this [Document Picker bug](https://github.com/Elyx0/react-native-document-picker/issues/202).

PR #687, opened 2019-06-09
Add module annotation

lithin:
In order to enable lazy package loading, we need an annotation for the native module.


itinance:
Many thanks!

lithin:
Thanks @itinance . When are you planning to do a release?

PR #693, opened 2019-06-19
Move the changelog to its own file

sbeca:


itinance:
Nice catch! Thanks a lot

PR #702, opened 2019-06-27
Fix: S3 pre signed uploads or similar binary stream uploads writing corrupt files

saurabytes:
## Background
RNFS is a great library to download and upload files. While trying to upload a file to S3 using a PreSigned `PUT` url, it was writing corrupt files as it was adding form boundary, meta data, etc to the file.

## Fix
`binaryStreamOnly`: A newly added field in `UploadFileOptions` which is defaulted to `false`.
If this options is set `true` the uploader for RNFS will only write file binary data to the directed upload url. It omits all the meta data and form boundaries as they will write to the uploading stream essentially writing a corrupt file for the format provided.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,5 @@ android/.gradle/*
 android/gradle/*
 android/*.iml
 android/local.properties
+android/.settings
+android/.project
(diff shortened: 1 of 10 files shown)


itinance:
This is a really good thing! Glad you find it out and fixed it!

PR #705, opened 2019-07-02
Fix #704

### LINKED ISSUE #704: Upload binary file corruption in Android
`uploadFiles()` method in android adds two `crlf` to the binary file one at the beginning and one at the end of the file and makes the uploaded file corrupted.

image

image

These images show the `0d0a` added to the file that is the `crlf` in hex.

HassanMojab:
I've fixed the problem by removing the two `crlf`s.


itinance:
Hey community, any thoughts on this?

Issue #443, opened 2018-02-16
Android downloadFile completely freezes the app


mdrbohlav:
When I start downloading large file the app completely freezes.

My config is really simple:
```
const config = {
  fromUrl: url,
  toFile: path,
  progressDivider,
  begin: ({ contentLength }) => {
    processProgress(contentLength);
  },
  progress: ({ contentLength, bytesWritten }) => {
    processProgress(contentLength, bytesWritten);
  },
};

const task = RNFS.downloadFile(config);
```

Android: 6.0 and 7.1.1, both emulated
RN: 0.51.0
RNFS: 2.9.10

itinance:
At zizzle we use this function to download large files and even show a progressbar at the bottom of the screen, while the user can navigate between different views without and freeze.

Seems that something different is causing your issue.

mdrbohlav:
There is nothing else happening in the moment of the download. But I found for some reason on Android the file size of a specific file is not known (did not check if the header 'content-length' is set but it works fine on iOS) which means the progressDivider has no effect and it sends progress notification each few ms.

Kendysond:
For me, removing my download progress function resolved the freeze on Android. I guess it's because redux keeps updating the state at each progress increment. Sad that android users can't see an actual progress view, but  I guess it's cool.

sergibel:
Option _progressDivider_ do not work if the server does not provide response content length. To reduce the frequency of callbacks, the option _progressInterval_ defines how often the progress is returned, see PR #717

PR #717, opened 2019-07-18
bugfix #443: Add option progressInterval for downloadFile

sergibel:
Hello,

Option _progressDivider_ do not work if the server does not provide response content length. To reduce the frequency of callbacks, the option _progressInterval_ defines how often the progress is returned.


itinance:
Many thanks for this @sergibel !

Issue #656, opened 2019-04-11
Download doesn't timeout on IOS


wamry:
I checked using network link conditioner with 100% loss, even with turning off the wifi, the download doesn't fail.
I like the resumable feature, but what if i want it to fail after the timeout is exceeded ?

This is my downloadFileOptions object:
```
const downloadFileOptions = { 
	fromUrl,
	toFile,
	background: true,
	discretionary: true,
	cacheable: false,
	progressDivider: 5,
	readTimeout: 15000,
	connectionTimeout: 5000,
	progress: (data) => {
		myCallBackOnProgress(data);
	},
	resumable:() => {
		return false
	}
}
```

a-koka:
It's a known issue that when iOS enables background downloading, it ignores the timeout errors. Just disable background downloading when timeouts are important.

@itinance Perhaps we should add this information to the documentation. I lost a half day trying to figure this out myself. Thank you!

wamry:
tried it with "background: false", still wouldn't timeout on ios.

a-koka:
Below is my code. Can you try with something similar? You have some extra props that may be getting in the way.

```
                RNFS.downloadFile({
                      fromUrl,
                      toFile,
                      progressDivider,
                      begin: res => {
                        contentLength = res.contentLength;
                        if (onBegin) onBegin(res);
                      },
                      progress: onProgress,
                      connectionTimeout: 1500,
                      readTimeout: 1500,
                      background: false
                 }).then().catch(err => // Get the timeout error here)
```

@itinance Forgot to mention earlier, the documentation should also say that the timeouts are in milliseconds. Thank you!

PR #725, opened 2019-08-01
Support iOS background download timeout

hojason117:
This PR adds the ability to timeout iOS background downloads.

Currently the only way to timeout iOS downloads is the `readTimeout` option, but this isn't working for background downloads. `readTimeout` uses [timeoutIntervalForRequest](https://developer.apple.com/documentation/foundation/nsurlsessionconfiguration/1408259-timeoutintervalforrequest?language=objc) under the hood, but since download tasks created by a background session are automatically retried (up to 7 days) if failed due to a timeout, the exception thrown by `readTimeout` is ignored in background mode.

By adding the `backgroundTimeout` option to `downloadFile`, which uses [timeoutIntervalForResource](https://developer.apple.com/documentation/foundation/nsurlsessionconfiguration/1408153-timeoutintervalforresource?language=objc#declarations) under the hood, now we have a way to timeout background downloads.

[`timeoutIntervalForRequest` vs `timeoutIntervalForResource`](https://github.com/Alamofire/Alamofire/issues/1266#issuecomment-221471947)

Fixes: #656

### DIFF
--- a/Downloader.h
+++ b/Downloader.h
@@ -20,7 +20,8 @@ typedef void (^ResumableCallback)();
 @property        bool discretionary;                          // Whether the file may be downloaded at the OS's discretion (iOS only)
 @property        bool cacheable;                              // Whether the file may be stored in the shared NSURLCache (iOS only)
 @property (copy) NSNumber* progressDivider;
-@property (copy) NSNumber* readTimeout;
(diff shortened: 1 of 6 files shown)


itinance:
Many thanks @hojason117!

Issue #727, opened 2019-08-05
iOS uploads stuck in binaryStreamOnly mode


ball-hayden:
On iOS, files are always uploaded in binaryStreamOnly mode.

I'm not an Objective C expert, but the issue seems to be that the assignment here isn't taking.

https://github.com/itinance/react-native-fs/blob/3bc8e393463022d1668af8d6075498d7310d58aa/RNFSManager.m#L591

```
(lldb) print options[@"binaryStreamOnly"]
(__NSCFBoolean *) $15 = 0x000000010eb98380 NO
(lldb) print params.binaryStreamOnly
(BOOL) $16 = YES
```

ball-hayden:
I'm wondering if this is because the `params` object isn't actually initialised, but is only allocated?
Would that make any sense?

PR #728, opened 2019-08-05
Correctly read binaryStreamOnly param

ball-hayden:
Fixes https://github.com/itinance/react-native-fs/issues/727

`options[@"binaryStreamOnly"]` isn't a `BOOL`, but is actually a `_NSCFBoolean`.
To correctly read the value of it, we must get the `boolValue` of the object.


itinance:
Many thanks!

PR #731, opened 2019-08-08
2.14.2

snowiosic:
解决 iOS端  自定义https 无法下载的问题


itinance:
Thanks for this. But I think we should enable this on behalf of an optional setting in downloadOptions. Do you agree?

PR #752, opened 2019-09-09
Fix Xcode and Java deprecation warnings

jnpdx:
This pull request does two things:

1. Fix simple syntax errors and method deprecations (no change in functionality)
2. Replaces the old deprecated `NativeAppEventEmitter` with `RCTEventEmitter`

For the first change, everything is pretty straightforward except for two changes I made to error handling -- in both cases.  RNFS was tacitly doing some dangerous/unworkable casts (`NSException` to `NSError`, `NSString` to `NSInteger`).  I fixed these, and made the error codes `0` in both cases.  If they should have different values, that should be changed before merging.

In order to accommodate the second change, I also had to change the event names, since iOS must supply an exhaustive list of supported events in the `supportedEvents` method.  So, I removed the `jobId` from the end of the event name since it's used in the body info of the event anyway.  In my testing, this doesn't effect multiple simultaneous uploads/downloads in practice.

Deployment target was moved to 8.0 since lots of code was using functionality introduced in 8.0 and thus would've failed anyway on a 7.0 target.

Using Xcode 10.2.1, the project is now free of warnings.

On the Android side, I moved the minimum API version up to 19, since the project was using a method not available until 19 (in conflict with the original min API of 16).  This should leave the Android side without warnings and errors as well.


itinance:
MAny thanks for this! Great work! I will add a specific comment to the event name section.

PR #754, opened 2019-09-10
iOS: Fix binaryStreamOnly not properly being converted from NSNumber to BOOL

LowieHuyghe:
iOS: Fix binaryStreamOnly not properly being converted from NSNumber to BOOL
binaryStreamOnly was always true


itinance:
Thanks for this. Already merged today by #728. Thanks again :)

Issue #749, opened 2019-09-06
RNFS.uploadFiles upload raw not multipart


jsina:
I've tried to upload file from RNFS from iPad air2, but when I log it out from `nc -l` I see that there is no `Content-Disposition` and `Content-Type`,  in fact it seems to be raw instead of multipart.

```
POST / HTTP/1.1
Host: 192.168.1.5:10002
Content-Type: multipart/form-data; boundary=----A63774C3-4BB0-499B-975A-43D99F8D07B4
Connection: keep-alive
Accept: application/json
User-Agent: DNA/1 CFNetwork/978.0.7 Darwin/18.2.0
Content-Length: 711529
Accept-Language: en-us
Accept-Encoding: gzip, deflate
```
I'm using ```"react-native-fs": "^2.13.3",``` and this is my react-native info:
```
System:
    OS: macOS 10.14.3
    CPU: (4) x64 Intel(R) Core(TM) i5-6267U CPU @ 2.90GHz
    Memory: 64.70 MB / 8.00 GB
    Shell: 2.6.0 - /usr/local/bin/fish
  Binaries:
    Node: 10.16.0 - ~/.config/fnm/bin/node
    Yarn: 1.13.0 - /usr/local/bin/yarn
    npm: 6.9.0 - ~/.config/fnm/bin/npm
    Watchman: 4.7.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 12.4, macOS 10.14, tvOS 12.4, watchOS 5.3
    Android SDK:
      API Levels: 16, 21, 23, 25, 26, 27, 28
      Build Tools: 23.0.1, 25.0.0, 25.0.1, 25.0.2, 25.0.3, 26.0.0, 26.0.1, 26.0.2, 26.0.3, 27.0.0, 27.0.1, 27.0.2, 27.0.3, 28.0.3
      System Images: a...ult | ARM EABI v7a, a...-21 | Intel x86 Atom, a... | Intel x86 Atom_64
  IDEs:
    Xcode: 10.3/10G8 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.8.6 => 16.8.6 
    react-native: ^0.60.5 => 0.60.5 
  npmGlobalPackages:
    create-react-native-app: 2.0.2
    react-native-cli: 2.0.1
    react-native-git-upgrade: 0.2.7
```

hank121314:
[...]

PR #759, opened 2019-09-17
Fix for issue #749: RNFS.uploadFiles upload raw not multipart

scottewart:
The RCT_EXPORT_METHOD for uploadFiles was casting the config option of binaryStreamOnly to it's internal params. The option comes through as an NSNumber as with all boolean primitives via the RCT methods. In objective-c comparing a NSNumber to a BOOL directly will do a pointer comparison, meaning that it will always be true if anything is set (eg: 0, 1, 999, fizzbuzz, as long as it's an object, the pointer will be true). To fix this you need to use the boolValue method on NSNumber to return a BOOL that matches the original intent.

See: https://github.com/itinance/react-native-fs/issues/749

### DIFF
--- a/RNFSManager.m
+++ b/RNFSManager.m
@@ -588,7 +588,7 @@ RCT_EXPORT_METHOD(uploadFiles:(NSDictionary *)options
   NSNumber* jobId = options[@"jobId"];
   params.toUrl = options[@"toUrl"];
   params.files = options[@"files"];
-  params.binaryStreamOnly = options[@"binaryStreamOnly"];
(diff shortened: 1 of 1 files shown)


scottewart:
Hi @itinance, please take a look at my PR, it fixes a bug that only affects iOS multipart fileUploads.

craftmusic:
👍 bless you, uploading files on ios was truly broken for me, the file upload request was indeed malformed due to this bug

This PR fixed the issue and my server is now able to detect the files as part of the request.

itinance:
Thanks for this!

scottewart:
No problem

PR #779, opened 2019-10-10
Add conditional comments around methods not supported in Mac Catalyst

leonid-shevtsov:
I've successfully built and used `react-native-fs` on macOS 10.15 with the just-released Mac Catalyst. 🎉 

There was one slight issue: a couple of methods rely on deprecated iOS APIs (see https://developer.apple.com/documentation/photokit/phasset/1624782-fetchassetswithalasseturls). These APIs aren't supported under Mac Catalyst.

I don't know if it's worth fixing the methods (haven't used them myself), but at least there's an easy way to exclude them from the Catalyst build with conditional comments.

Thanks!

### DIFF
--- a/RNFSManager.m
+++ b/RNFSManager.m
@@ -708,6 +708,8 @@ RCT_EXPORT_METHOD(getFSInfo:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromise
 }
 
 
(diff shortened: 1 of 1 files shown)


itinance:
Thanks! I've added notes on the README that these functions won't be available on Mac Catalyst

PR #790, opened 2019-10-21
Pass empty callbacks to emitters in the event of no parameters

### LINKED ISSUE #788: Another bug in 2.15
Issues a warning:
```
Sending `DownloadBegin` with no listeners registered.
```
when I do
```
  let {promise} = fs.downloadFile({fromUrl, toFile});
```

jnpdx:
Should fix #788 issue with event emitter warnings when there are no listeners by passing empty callbacks.


StasD:
Don't you also need to update `Upload...` counterparts in a similar way by any chance? I saw then in PR #752.

jnpdx:
@StasD good call.  Added

PR #793, opened 2019-10-24
Improve implementation related to event callbacks

ifsnow:
It's related to #790, #788

I think this implementation is better than passing empty callbacks.
In most cases where no events are required, works more efficiently on both Android and iOS.
As you know, performance is improved because unnecessary bridge communication is reduced.

I hope this helps.


itinance:
WOW! Thanks for this!

Issue #791, opened 2019-10-22
Build error: 'React/RCTImageLoaderProtocol.h' file not found for react native version 0.56


adkaushik:
The recent changes to the repo to fix the issue with RCTImageLoader for react native version >= 0.60, are failing builds in version 0.56.

RNFS Version: 2.16.1. 

Should we just fix with previous version of the library?

perfectAlways1028:
I got that error when I try with react-native version 0.60.6.

RNFS Version: 2.14.1

NobodyFlex:
Same issue
RN: 0.59.9
RNFS: 2.14.1

itinance:
As written in README: for RN < 0.61, install 2.15
and for RN >= 0.61, install 2.16

AnthonyAkentiev:
@itinance 
RN: 0.60.5
RNFS: 2.15.0

You mentioned above that i should use 2.15, but i am still getting this error :-(((
Please help

grit96:
The README says for RN >= 0.57 install RNFS >= 2.13.2, and for RN >= 0.61 install RNFS >= 2.16.0.

It doesn't say that RNFS 2.16.x is not compatible with RN < 0.61.
If that is the case, then the README needs updating to be clearer.

AnthonyAkentiev:
@itinance @grit96 
2.15.2 worked well
2.15 didn't

itinance:
@AnthonyAkentiev are you sure that react-native-fs 2.15.x is installed?

Please double check the `include`-statement in line 17 at `node_modules/react-native-fs/RNFSManager.m`

AnthonyAkentiev:
@itinance Please see my prev.comment above.
2.15.2 worked well, thank you!

AlanKeiss:
Changing 
`#import` <React/RCTImageLoaderProtocol.h>`
to
`#import <React/RCTImageLoader.h>`
fixed the build issue for me

mattslight:
For RN 0.60.5 use 2.15.2
For RN 0.61 use 2.16

viventus:
If you are breaking API then you must up MAJOR version! whats wrong with you?

luatnd:
Instead of changing `#import` statement again again and again, why don't you use a suitable version like @mattslight said !!

esutton:
Updating to 2.16.1 broke my build .Rolling back to 21.15.2 fixed it. @mattslight thanks!
````
"react": "^16.9.0",
[...]

PR #806, opened 2019-11-07
Update readme for RN 0.59 usage

vforvasile:
Fixes iOS build [https://github.com/itinance/react-native-fs/issues/791](https://github.com/itinance/react-native-fs/issues/791)


grit96:
I believe this is now fixed in bbafdc1 as of v2.16.4

Issue #700, opened 2019-06-26
`writeFile` on Android Simulator will not overwrite an existing file, but write over it


pedrosimao:
I am using the `RNFS.writeFile(path, myObject, 'utf8');` method to write on a file specified by `path`. It works perfectly on iOS, but on Android Simulator it behaves strangely. When I use the command to write on a file that already exists, it will not overwrite the existing file, but it will write over it. The consequence is that if the existing file has a longer string than the string I am trying to write, the end of the string will be a broken thing, very weird.

I have tried to do `exists()` and `unlink()` in case the file exists before writing and it eliminates the bug. But the app becomes slower and if the user writes too fast I get some other annoying bugs because file does not exists when I try to read it. 

Did anyone else had the same trouble as me?

pedrosimao:
Finally, I solved the issue by ditching `react-native-fs` and using `rn-fetch-blob` instead. I couldn't afford to wait for the issue in this lib to be fixed. Hope the issue will be investigated anyway.

gpawlik:
@pedrosimao I am experiencing a similar issue but only on Android Q (API 29), on older versions it looks fine, could you tell which version were you using, so we could narrow down the problem?

hammadj:
@pedrosimao @gpawlik Just running into this now as well on Android Q. Any luck with it?

sbaruth:
Occurring for me as well. Pixel 2, Android 10, physical device.

rogerm4242:
I am not a developer, but this sounds like the same behaviour I see in the Joplin Android app when using local storage sync on Android 10.

Whenever I delete information from a note, the same number of characters stay appended to the end of the file, causing the sync process to fail.

I have an original Pixel XL and a Pixel 3a XL, both running Android 10.

mstratiev:
Same issue here with production app, all devices running Android 10 - not writing over the whole file, but only the part that overlaps - i.e. if I want to write 100 symbols to a 120-symbol file, only the first 100 are overwritten and the rest remain, which in my case break the app. Devices include Samsung Note-s, OnePlus, Nokia etc. all that have A10 updated. Might be a change in the the Android 10 API that is used in this package.

The fix that I used, because I really needed to fix the production was - with each write, first delete the file:

RNFS.unlink(path).then(() => {
  RNFS.writeFile(path, content)); 
});

mstratiev:
[...]

PR #837, opened 2020-01-15
writeFile() not truncating file contents on Android 10

SilverInfinity:
Fix for #700 
As of Android 10, passing just "w" will no longer truncate the file after writing. You must also explicitly pass the truncate option. Previous versions of android treated "w" and "wt" as the same.

Changes:
- send write/truncate option as the mode to openOutputStream on Android when append is false.

### DIFF
--- a/android/src/main/java/com/rnfs/RNFSManager.java
+++ b/android/src/main/java/com/rnfs/RNFSManager.java
@@ -118,7 +118,7 @@ public class RNFSManager extends ReactContextBaseJavaModule {
     Uri uri = getFileUri(filepath, false);
     OutputStream stream;
     try {
-      stream = reactContext.getContentResolver().openOutputStream(uri, append ? "wa" : "w");
+      stream = reactContext.getContentResolver().openOutputStream(uri, append ? "wa" : "wt");
     } catch (FileNotFoundException ex) {
       throw new IORejectionException("ENOENT", "ENOENT: " + ex.getMessage() + ", open '" + filepath + "'");
     }


Iyoss:
@SilverInfinity hello.
Are you sure that we can use "wt" mode? I'm asking because I can't find this mode in Android [documentation](https://developer.android.com/reference/android/content/ContentResolver?hl=es-419.COMOacute%20PONER%20CARATULAS%20A%20TUS%20CANCIONES%20FAacuteCIL%20NO%20ROOTCOMO%20USAR%20FACEBOOK%20OFICIAL%20SIN%20MESSENGER%20EN%20ANDROID%202016FAacuteCIL%20NO%20ROOTABREMEEE!!!!#openOutputStream(android.net.Uri,%20java.lang.String)). But your fix works.

yohannpoli:
Any news on that issue please ? Can you please merge it?

SilverInfinity:
@Iyoss 
I came across it while trying to find more detail on the modes. 
ContentResolver.openOutputStream creates an AssetFileDescriptor using [openAssetFileDescriptor](https://developer.android.com/reference/android/content/ContentResolver?hl=es-419.COMOacute%20PONER%20CARATULAS%20A%20TUS%20CANCIONES%20FAacuteCIL%20NO%20ROOTCOMO%20USAR%20FACEBOOK%20OFICIAL%20SIN%20MESSENGER%20EN%20ANDROID%202016FAacuteCIL%20NO%20ROOTABREMEEE!!!!#openAssetFileDescriptor(android.net.Uri,%20java.lang.String)), then calls openOutputStream on that.
The openAssetFileDescriptor method's mode parameter references [ContentProvider#openAssetFile](https://developer.android.com/reference/android/content/ContentProvider?hl=es-419.COMOacute%20PONER%20CARATULAS%20A%20TUS%20CANCIONES%20FAacuteCIL%20NO%20ROOTCOMO%20USAR%20FACEBOOK%20OFICIAL%20SIN%20MESSENGER%20EN%20ANDROID%202016FAacuteCIL%20NO%20ROOTABREMEEE!!!!#openAssetFile(android.net.Uri,%20java.lang.String)) for available parse modes.
Here, I found a bit more description: 

>mode: 	String: Access mode for the file. May be "r" for read-only access, "w" for write-only access (erasing whatever data is currently in the file), "wa" for write-only access to append to any existing data, "rw" for read and write access on any existing data, and "rwt" for read and write access that truncates any existing file. This value must never be null.

However, this appears to not be correct, as on android 10, passing "w" for write-only access is not in fact erasing whatever data is currently in the file.

This method linked out to [ContentProvider.openFile](https://developer.android.com/reference/android/content/ContentProvider?hl=es-419.COMOacute+PONER+CARATULAS+A+TUS+CANCIONES+FAacuteCIL+NO+ROOTCOMO+USAR+FACEBOOK+OFICIAL+SIN+MESSENGER+EN+ANDROID+2016FAacuteCIL+NO+ROOTABREMEEE%21%21%21%21#openFile(android.net.Uri,%20java.lang.String,%20android.os.CancellationSignal)) which mention's [ParcelFileDescriptor.parseMode(String)](https://developer.android.com/reference/android/os/ParcelFileDescriptor?hl=es-419.COMOacute+PONER+CARATULAS+A+TUS+CANCIONES+FAacuteCIL+NO+ROOTCOMO+USAR+FACEBOOK+OFICIAL+SIN+MESSENGER+EN+ANDROID+2016FAacuteCIL+NO+ROOTABREMEEE%21%21%21%21#parseMode(java.lang.String))
Its here that I found all of the modes listed:
> mode	String: The string representation of the file mode. Can be "r", "w", "wt", "wa", "rw" or "rwt".

That, coupled with this [android bug ticket](https://issuetracker.google.com/issues/135714729), which mention's needing to pass 't' in the open mode is where I got the "wt" file mode.
I poked around the Android source code for different versions, and found that before Android 10, "w" and "wt" were explicitly being parsed exactly the same in ParcelFileDescriptor, resulting in a bitmask with both MODE_WRITE_ONLY and MODE_TRUNCATE to open the file. Then in Android 10, there was a change and 'w' and 't' are now parsed individually.

To answer your question, looking back on it, I am not _entirely_ sure. The documentation on its own did not appear to be entirely clear on it, or contradictory in places. I am sure that it works for absolute file paths, such as files stored in RNFSDocumentDirectoryPath as is the use case for my application where i noticed the issue. Also tested on multiple versions of android. However I haven't tested it with content uris. If it is a content uri, then it looks like the Content Resolver or Content Provider could be overwritten with a custom implementation? I am not sure what to do in such a case.

I can change "wt" to "rwt". "rwt" also fixes the issue, and is explicitly mentioned in the android documentation. However, that may require adding READ_EXTERNAL_STORAGE permissions where it was previously not required. I was trying to avoid that.

SilverInfinity:
As I have explained above, Would you like me to edit this PR to use "rwt" or leave it as "wt"? Both options solve the issue.

I have been using a forked branch in production for 2 months now, and have seen no issues in using "wt". However our use case is very simple, only writing txt and image files to the document directory.
I don't have the time to test other scenarios myself. Since "rwt" is explicitly mentioned in the documentation, if you would rather use that, then I will update this PR.

itinance:
Anything we can do here to find a final solution for your discussion, guys?

creambyemute:
@itinance I think the solution proposed here and in https://github.com/itinance/react-native-fs/pull/890 fits.

In our project we now have to use forks of redux-persist-fs-storage and react-native-fs due to the Android 10 problem. Otherwise itt pretty much rendered our application unusable.

JordanDuncan:
I can verify that this fixes an issue I was having on an app that was using redux-persist-fs. 

Any time a reducers content was reduced in size and commited to disk on android 10, the app would be unable to rehydrate any data for that reducer as it left non-valid JSON in the persistors file.

This patch has fixed the android 10 issue and works as expected on previous versions (physically tested on 9 and 8).

exotexot:
I also would like to see this merged

gnardini:
Just one more person sharing that they need to use 2 forks because of this, please merge 😄

sh-zam:
Hello @SilverInfinity 

Disclaimer: I found this issue while researching. I don't use this library or react native. 

> As I have explained above, Would you like me to edit this PR to use "rwt" or leave it as "wt"?

We used "wt", after sometime users reported that they can no longer save the files. This specifically happen to files from Google Drive. The error we were getting is: `java.io.FileNotFoundException: Unsupported mode: wt`. The problem was fixed by switching to "rwt". 

Again, I don't know this library but I felt like reporting it as it seemed relevant :)

ashughes:
@sh-zam Thanks for the heads up about `"wt"` not working with Google Drive. I was able to reproduce this as well.

I just [filed a bug for this on the Android issue tracker](https://issuetracker.google.com/180526528). Hopefully, someone will follow up with more guidance and updated documentation.

FYI, I'm an Android developer, not a react native developer or user of this library. I also got here while researching this issue.

FeXd:
I just spent waaaaaaay too much on a bug caused by this. Can we please have an update on gettings this approved?

gaodeng:
any news?

Issue #811, opened 2019-11-15
Concurrent downloads send progress updates to all progress callbacks


respectTheCode:
When 2 concurrent downloads are running the progress callback for each download will be called for progress from either download.

The if statement in the progress callback bellow should never be true.

```
const fileDownload = fs.downloadFile({
	fromUrl,
	toFile,
	progress: progress => {
		if (progress.jobId != fileDownload.jobId) {
			console.log(
				"jobId mismatch",
				progress.jobId,
				fileDownload.jobId
			);
			return;
		}
```

This happens because the concurrent downloads all broadcast the same `DownloadProgress` event across the bridge.

creambyemute:
Maybe related:

If I have multiple concurrent file downloads (for signature picutres, pictures and pdfs) with the following code:

```javascript
const result = RNFS.downloadFile({
    fromUrl: `${Constants.DatafilesApi.GET_FILE_ONLY_ENDPOINT}${file.id}`,
    toFile: `${signaturesPath}${file.id}.${file.extension}`,
    headers: {
        Authorization: token
    },
    discretionary: true,
    cacheable: true
});
const response = await result.promise;
```

It sometimes writes the content to the wrong file/destination (for example the signature is written to a picturePath instead of desired signaturePath.

When swapping RNFS.downloadFile with RNFetchBlob on the other hand, it works as expected, the images/signature are always saved to the correct path with the correct name.

cinjon:
Confirmed that this is an issue.

itinance:
Thanks to #842, I have published 2.16.4 with the fix for this.
I am going to close this issue, but pls feel free to reopen if the issue persists

MaxInMoon:
Same issue on 2.16.6 for me on Android, I am the only one?

PR #842, opened 2020-02-03
add condition inside download event, before pass it to event callback.

RZulfikri:
Related issue: https://github.com/itinance/react-native-fs/issues/811

**Problem**
When I called multiple `RNFS.downloadFile`, `begin` and `progress` callbacks return all events that occur, not only event that related to that instance only.

For example, I have 2 download process with jobId 1 and 2. In `begin` callback I'll receive responses of jobId 1 and 2, also in the `progress` callback.

**Solution**
When I check the native code there is nothing wrong with that, so I add conditions in event listener to check jobId before pass it event callback.

```javascript
    if (options.begin) {
      subscriptions.push(RNFS_NativeEventEmitter.addListener('DownloadBegin', (res) => {
        if (res.jobId === jobId) options.begin(res);
      }));
    }

    if (options.progress) {
      subscriptions.push(RNFS_NativeEventEmitter.addListener('DownloadProgress', (res) => {
        if (res.jobId === jobId) options.progress(res);
      }));
    }

    if (options.resumable) {
      subscriptions.push(RNFS_NativeEventEmitter.addListener('DownloadResumable', (res) => {
        if (res.jobId === joibId) options.resumable(res);
      }));
    }
```

### DIFF
--- a/FS.common.js
+++ b/FS.common.js
@@ -500,15 +500,21 @@ var RNFS = {
     var subscriptions = [];
 
     if (options.begin) {
-      subscriptions.push(RNFS_NativeEventEmitter.addListener('DownloadBegin', options.begin));
+      subscriptions.push(RNFS_NativeEventEmitter.addListener('DownloadBegin', (res) => {
+        if (res.jobId === jobId) options.begin(res);
+      }));
     }
 
     if (options.progress) {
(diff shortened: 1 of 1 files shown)


itinance:
Many thanks!

PR #845, opened 2020-02-06
Fix typo (joib !== job)

willfulbard:


willfulbard:
I realized that this fix is not correct, I'm putting together a better PR soon.

PR #846, opened 2020-02-06
Fix flow errors and typo

willfulbard:
- Flow is expecting that the `resumable` callback doesn't accept arguments.  As the event sent by the native code doesn't send a payload either, this should be reflected here.
- Flow is complaining about the callbacks possibly being undefined.  This is because, as far as flow knows, between the time that the callback is set-up and the callback is fired the callback could be undef (as is according to the flowtype definition).  Therefore, it needs to be checked at BOTH at the time the callback is setup, and at the time the callback is fired.
- `joibId` is a typo, but is not needed since there isn't ever a jobId in the resumable callback.

NOTE: I'm using versions:
```
"flow-bin": "~0.105.0",
"flow-typed": "^2.5.1",
"react-native": "0.61.5",
```

### DIFF
--- a/FS.common.js
+++ b/FS.common.js
@@ -501,19 +501,19 @@ var RNFS = {
 
     if (options.begin) {
       subscriptions.push(RNFS_NativeEventEmitter.addListener('DownloadBegin', (res) => {
-        if (res.jobId === jobId) options.begin(res);
+        if (res.jobId === jobId && options.begin) options.begin(res);
       }));
     }
 
     if (options.progress) {
       subscriptions.push(RNFS_NativeEventEmitter.addListener('DownloadProgress', (res) => {
-        if (res.jobId === jobId) options.progress(res);
+        if (res.jobId === jobId && options.progress) options.progress(res);
       }));
     }
 
     if (options.resumable) {
-      subscriptions.push(RNFS_NativeEventEmitter.addListener('DownloadResumable', (res) => {
-        if (res.jobId === joibId) options.resumable(res);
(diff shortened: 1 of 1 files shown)


FS.common.js
```diff
@@ -501,19 +501,19 @@ var RNFS = {
...
-        if (res.jobId === jobId) options.progress(res);
+        if (res.jobId === jobId && options.progress) options.progress(res);
       }));
     }
 
     if (options.resumable) {
-      subscriptions.push(RNFS_NativeEventEmitter.addListener('DownloadResumable', (res) => {
-        if (res.jobId === joibId) options.resumable(res);
+      subscriptions.push(RNFS_NativeEventEmitter.addListener('DownloadResumable', () => {
+        if (options.resumable) options.resumable();
```
dstop75: Just noticed this PR after submitting #852 to only call resumable for the correct jobId, maybe they should be combined? I'm not familiar with flow so I didn't notice or address that concern...

grit96:
This would be really useful to get merged. Currently I'm having to exclude `react-native-fs` from flow checking.

FS.common.js
dstop75: Just wondering if this type of approach would satisfy flow with fewer checks?
```javascript
    var progress = options.progress;

    if (progress) {
      subscriptions.push(RNFS_NativeEventEmitter.addListener('DownloadProgress', (res) => {
        if (res.jobId === jobId) progress(res);
      }));
    }
```

PR #848, opened 2020-02-09
Add support for ph:// uris to copyAssetsVideoIOS

### LINKED ISSUE #805: copyAssetsVideoIOS does not work with ph://
I'm having an issue where my code i just hanging when i call copyAssetsVideoIOS, as well as copyAssetsFileIOS with the URI given back to me via react-native-cameraroll.

Any help would be amazing

Here is the code i am using

```
const medias = await Promise.all(
            selected.map(async index => {
                let media = photos.edges[index].node
                let ext = media.image.filename.split('.')[1];
                if(Platform.OS === 'ios') {
                    const dest = `${RNFS.TemporaryDirectoryPath}${Math.random().toString(36).substring(7)}.${ext}`;
                    let copiedPath = ''
                    
                    if(media.type === 'video') {
                        copiedPath = await RNFS.copyAssetsVideoIOS(media.image.uri, dest);
                    }
                    else if(media.type === 'image') {
                        copiedPath = await RNFS.copyAssetsFileIOS(
                            media.image.uri,
                            dest,media.image.width,
                            media.image.height,
                            1.0,
                            1.0,
                            'contain'
                        );
                    }
                    else {
                        throw new Error("Invalid media type, only image and videos are acceptable")
                    }

                    const stats = await stat(copiedPath);
                    return { ...stats, uri: media.image.uri };
                }
                else {
                    // Android
                    const stats = await stat(media.image.uri);
                    return { ...stats, uri: media.image.uri };
                }
            })
        )
```
Variables
photos.edges are the edges from the cameraroll documentation

the URI come back as a ph:// prefix

Thanks!

enumura:
resolves #805

### DIFF
--- a/RNFSManager.m
+++ b/RNFSManager.m
@@ -860,7 +860,13 @@ RCT_EXPORT_METHOD(copyAssetsVideoIOS: (NSString *) imageUri
   //__block NSURL* videoURL = [NSURL URLWithString:destination];
   __block NSError *error = nil;
 
-  PHFetchResult *phAssetFetchResult = [PHAsset fetchAssetsWithALAssetURLs:@[url] options:nil];
+  PHFetchResult *phAssetFetchResult = nil;
+  if ([url.scheme isEqualToString:@"ph"]) {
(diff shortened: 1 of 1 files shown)


itinance:
Thank you very much!

PR #852, opened 2020-02-18
Add fix and improvements for downloadFile event handlers:

dstop75:
Invoke resumable only for the correct jobId

  Throw errors if defined but not functions

  Remove subscriptions on both fulfilled/rejected


FS.common.js
```diff
@@ -535,13 +538,9 @@ var RNFS = {
...
     return {
       jobId,
-      promise: RNFSManager.downloadFile(bridgeOptions).then(res => {
+      promise: RNFSManager.downloadFile(bridgeOptions).finally(() => {
         subscriptions.forEach(sub => sub.remove());
-        return res;
       })
-        .catch(e => {
-          return Promise.reject(e);
-        })
```
dstop75: Note: Subscriptions are now removed when `downloadFile` rejects, in addition to when it fulfills. But if the download is stopped and never resumed, they are still not removed from the EventEmitter, at least on iOS. I don't see a way to say "abandon this job and clean up all the listeners", am I missing something? Stale subscriptions will no longer be fired here due to the `jobId` checks, but I think they will still build up and leak some small amount of memory, won't they?

PR #859, opened 2020-02-29
Added Android support for symlinking

crutchcorn:
I found a usecase where symlinking would be useful to have. This adds support for Android using symlink functionality


crutchcorn:
Just added support for iOS as well. Confirmed that it builds and works, as it is being utilized in production for one my applications

PR #860, opened 2020-03-02
Update package.json

FDiskas:
Fix: `react-native-fs@2.16.6" has incorrect peer dependency "react-native@^0.59.5".`

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "prop-types": "^15.7.2"
   },
   "peerDependencies": {
-    "react-native": "^0.59.5",
+    "react-native": "^0.59.5 || ^0.61.5",
     "react-native-windows": "^0.57.2"
   }
 }


crutchcorn:
Neither this PR, nor #832 are updating the peer deps to what they should be. `||` indicates an "or", not a range. We should be looking at something like:

```json
"react-native": ">=0.59.5 <0.64.0"
```

Which would allow anything from `0.58.5` to `0.64.0` to install

That said, @itinance this PR is increasingly important as npm 7 breaks installs (throws hard error) with incorrect peer deps

package.json
```diff
@@ -35,7 +35,7 @@
     "prop-types": "^15.7.2"
   },
   "peerDependencies": {
-    "react-native": "^0.59.5",
+    "react-native": "^0.59.5 || ^0.61.5",
```
crutchcorn, line 38:
```suggestion
    "react-native": ">=0.59.5 <0.64.0",
```

crutchcorn (approved):

package.json
```diff
@@ -35,7 +35,7 @@
     "prop-types": "^15.7.2"
   },
   "peerDependencies": {
-    "react-native": "^0.59.5",
+    "react-native": ">=0.59.5 <0.64.0",
     "react-native-windows": "^0.57.2"
```
crutchcorn, line 39: Actually, it's worth mentioning that this should be updated with the same expression as above.

package.json
```diff
@@ -35,7 +35,7 @@
...
   }
```
crutchcorn, line 40:
```suggestion
  },
  "peerDependenciesMeta": {
    "react-native-windows": {
      "optional": true
    }
  }
```

This will make sure that npm 7 does not automatically install `react-native-windows` when installing this package. Very important

package.json
crutchcorn:
```suggestion
    "react-native-windows": ">=0.59.5 <0.64.0"
```

crutchcorn:
After these last two changes, everything should be good-to-go in terms of merging

FDiskas:
and... later this year :)

PR #872, opened 2020-03-31
Binary only stream file length regression

stobob:
This fixes an issue with progress update events not matching the jobId.  And a regression that occurred when the binaryonlystream option was added on Android.  The request length was changed to being set only when in not in binarystream mode. This caused the progress events to pile up waiting for the file to completely upload and large file size uploads to fail uploading entirely.

### DIFF
--- a/android/src/main/java/com/rnfs/Uploader.java
+++ b/android/src/main/java/com/rnfs/Uploader.java
@@ -117,11 +117,11 @@ public class Uploader extends AsyncTask<UploadParams, int[], UploadResult> {
             if (mParams.onUploadBegin != null) {
                 mParams.onUploadBegin.onUploadBegin();
             }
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/rnfs/Uploader.java
```diff
@@ -117,11 +117,11 @@ private void upload(UploadParams params, UploadResult result) throws Exception {
             if (mParams.onUploadBegin != null) {
                 mParams.onUploadBegin.onUploadBegin();
             }
+            long requestLength = totalFileLength;
+            connection.setFixedLengthStreamingMode((int)requestLength);
```
matt-oakes, line 121: This needs to be after the `if` statement to avoid issues with non-binary only uploads.

Issue #628, opened 2019-02-11
Support for NSApplicationSupportDirectory in iOS


andrewmarmion:
Would it be possible to add support for storing files in the `NSApplicationSupportDirectory`? 

It looks like it could be quite a simple change to the `RNFSManager.m` by adding a new constant to export.

` @"RNFSApplicationSupportDirectory": [self getPathForDirectory:NSApplicationSupportDirectory],`

PR #886, opened 2020-05-20
fix: add application support path for ios

vdmtrv:
## Summary

Adding the Application Support directory path to RNFS (addresses https://github.com/itinance/react-native-fs/issues/628)


BatDroid:
is there any issue on this pr that is not merged?

PR #901, opened 2020-07-01
Handle filename with spaces

songsterq:
On iOS, when the file path contains spaces, such as `file://abc/xyz%20123.jpg`, the normalize logic will just strip the first 7 characters and return `/abc/xyz%20123.jpg`. However, it should be`/abc/xyz 123.jpg` since the original file path is URI encoded. So we URI decode the file path before stripping.

### DIFF
--- a/FS.common.js
+++ b/FS.common.js
@@ -27,7 +27,7 @@ var getJobId = () => {
   return jobId;
 };
 
-var normalizeFilePath = (path: string) => (path.startsWith('file://') ? path.slice(7) : path);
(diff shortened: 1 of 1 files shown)


yaeda:
This improvement is also effective for files with multi-byte file names such as Japanese.

@itinance 
Would you like to merge & release?

RZulfikri:
I test this PR, looks good. Thanks @songsterq you help me. 👍🏼

traxx10 (approved):

PR #931, opened 2020-09-18
Replace requestb with hookbin (requestb.in is down)

retyui:


retyui:
cc @itinance, @keremoge, @maheshwarimrinal

retyui:
another polite reminder @itinance

PR #945, opened 2020-10-27
Add condition inside upload event before passing it to the event call…

Laurencarne:
…back.

Similar issue: [#811](https://github.com/itinance/react-native-fs/issues/811)
Similar issue found for the `downloadFile`, `begin` and `progress` callbacks, our issue was involving the `uploadFile` event however very similar logic has been applied for this solution.

Problem
When calling `uploadFile` multiple different times, the `begin` and `progress` callbacks return for all events and not just the event they relate too. 

Solution
Added a condition inside the event listener to validate the `jobId` before passing the event callback.

### DIFF
--- a/FS.common.js
+++ b/FS.common.js
@@ -564,17 +564,25 @@ var RNFS = {
     if (options.method && typeof options.method !== 'string') throw new Error('uploadFiles: Invalid value for property `method`');
 
     if (options.begin) {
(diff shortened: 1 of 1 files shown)


seancheung:
Same here. This should be merged

seancheung:
Start a new uploading task right after aborting the previous one will also triggers messed-up callbacks

k1mmm:
Same experience here, would be awesome if this could be merged @itinance =)

PR #952, opened 2020-11-14
New Implementation of RNFS for Windows

avmoroz:
Created new implementation of RNFS for Windows using C++.
Previous implementation is located in the windows_legacy folder. 

copyFolder implemented to account for slow file transfers.


exotexot:
Amazing! :)

itinance:
Thank you very much! How should we proceed here? I'd like to have at least one or two other devs testing this PR before merging. However, I don't have any Windows machine. So i definitely need support here :)

stmoy:
@itinance - I'll poke the RN folks at Microsoft to get some more eyes on this!

exotexot:
Using this PR for about a now. Using downloadFile and uploadFile including authorisation Headers, stat, readDir, writeFile, and read and copyFile across all platforms, iOS, android, macOS (catalyst) and Windows and it’s working as expected! Great job!

FS.common.js
```diff
@@ -16,6 +16,7 @@ var RNFS_NativeEventEmitter = new NativeEventEmitter(RNFSManager);
 var base64 = require('base-64');
 var utf8 = require('utf8');
 var isIOS = require('react-native').Platform.OS === 'ios';
+var isWindows = require('react-native').Platform.OS === 'windows'; // To accomodate Windows
```
asklar, line 19:
```suggestion
var isWindows = require('react-native').Platform.OS === 'windows'; // To accommodate Windows
```

FS.common.js
```diff
@@ -598,18 +606,29 @@ var RNFS = {
     };
   },
 
-  touch(filepath: string, mtime?: Date, ctime?: Date): Promise<void> {
+  touch(filepath: string, mtime?: Date, ctime?: Date, Creation?: boolean): Promise<void> {
     if (ctime && !(ctime instanceof Date)) throw new Error('touch: Invalid value for argument `ctime`');
     if (mtime && !(mtime instanceof Date)) throw new Error('touch: Invalid value for argument `mtime`');
     var ctimeTime = 0;
-    if (isIOS) {
+    if (isIOS || isWindows) { // Modified to accomodate Windows
```
asklar, line 613:
```suggestion
    if (isIOS || isWindows) { // Modified to accommodate Windows
```

FS.common.js
```diff
@@ -598,18 +606,29 @@ var RNFS = {
...
       ctimeTime = ctime && ctime.getTime();
     }
-    return RNFSManager.touch(
-      normalizeFilePath(filepath),
-      mtime && mtime.getTime(),
-      ctimeTime
-    );
+    if(isWindows) {
```
asklar, line 616:
```suggestion
    if (isWindows) {
```

Consider linting these files

README.md
```diff
@@ -425,11 +428,19 @@ Write the `contents` to `filepath` at the given random access position. When `po
 
 Moves the file located at `filepath` to `destPath`. This is more performant than reading and then re-writing the file data because the move is done natively and the data doesn't have to be copied or cross the bridge.
 
+Note: Overwrites existing file in Windows.
+
+### `copyFolder(srcFolderPath: string, destFolderPath: string): Promise<void>`
+
+Copies the contents located at `srcFolderPath` to `destFolderPath`.
+
+Note: Windows only. This method is recommended when directories need to be copied from one place to another.
```
asklar, line 437: The second part of the sentence doesn't seem to add much... maybe describe instead why you would use this as opposed to (...)? is this about appcontainer restrictions?

README.md
```diff
@@ -425,11 +428,19 @@ Write the `contents` to `filepath` at the given random access position. When `po
...
+
 ### `copyFile(filepath: string, destPath: string): Promise<void>`
 
 Copies the file located at `filepath` to `destPath`.
 
-Note: On Android copyFile will overwrite `destPath` if it already exists. On iOS an error will be thrown if the file already exists.
+Note: On Android and Windows copyFile will overwrite `destPath` if it already exists. On iOS an error will be thrown if the file already exists.
```
asklar, line 443: it would be good/important to match API behavior across platforms but I realize that's a bigger issue than this PR

windows/RNFS.Tests/ReactNativeCxxTests/JsonJSValueReader.h
```diff
@@ -0,0 +1,42 @@
...
+// Licensed under the MIT License.
+
+#pragma once
+
+#include "JsonReader.h"
+#include "winrt/Microsoft.ReactNative.h"
+
+namespace winrt::Microsoft::ReactNative {
+
+struct JsonJSValueReader : implements<JsonJSValueReader, IJSValueReader> {
```
asklar, line 11: This seems like it's re-implementing JSON parsing... we should just use the Windows.Data.Json WinRT APIs directly instead

windows/RNFS/RNFSManager.h
```diff
@@ -0,0 +1,180 @@
...
+#include <winrt/Windows.Foundation.h>
+#include <winrt/Windows.Security.Cryptography.h>
+#include <winrt/Windows.Security.Cryptography.Core.h>
+#include <winrt/Windows.Web.Http.h>
+
+namespace Cryptography = winrt::Windows::Security::Cryptography;
+namespace CryptographyCore = winrt::Windows::Security::Cryptography::Core;
+namespace RN = winrt::Microsoft::ReactNative;
+
+struct CancellationDisposable
```
asklar, line 16: make this and other structures/classes `final` whenever possible

asklar (changes requested):
main thing is to either remove the custom json parsing or please explain why we need to have custom parsing :)

README.md
avmoroz: the reason copyDirectory was added was because using copyFile to copy over the contents of a folder would take a long long time. It would take 40 seconds to copy over a 60 MB directory with a large number of files using copyFile recursively when it could just be copy-pasted within a fraction of a second in file explorer. It appeared that a lot of conversions needed to be done between calling a function via JavaScript to it being executed, and I am not as familiar with the inner workings of React Native for Windows as I should be. This method is a workaround that allows folders like that 60 MB directory mentioned earlier to be copied much faster than they would be without it.

asklar: Thanks @avmoroz! this is good feedback for the nodejs folks - we have definitely heard (and seen) nodejs is a lot slower on Windows than linux. I filed https://github.com/nodejs/node/issues/36158 - if you have any more data or can leave more context on that bug, that'd be super appreciated too

README.md
avmoroz: I agree, but it is outside of the scope of this PR for the time being.

windows/RNFS/RNFSManager.h
avmoroz: Added the final keyword here.

emmanuj:
Noob question here but would this work with auto-linking for windows out of the box? And is it possible to use the API directly in C# without any additional configuration?

creambyemute:
@avmoroz This is 🔥! Gonna try it tomorrow, I'm atm re-initialiasing RN-Windows for our App and not much is missing, Kudos.

itinance:
@asklar thank you very much for the review! As I can see, @avmoroz did many changes according to your appreciated comments.  How can we proceed here?

asklar:
Thanks @avmoroz and @itinance I'll take another look in the next couple of days/early next week

creambyemute:
@itinance Question about the `.exists` method on android/ios:
Does it behave the same on iOS/Android when passing (the only difference is the missing / at the end):
* `startPath/somePath/`
* `startPath/somePath`

I'm asking because I had to modify my `.exists` calls for folders for RN-Windows (@asklar @avmoroz ):
(Path seperators are set correctly)
```javascript
if (isWindows) {
    const directoryExists = await this.RNFS.exists(targetDirectory.slice(0, -1)); // Remove the trailing slash, otherwise it woudl always yield false even if the directory exists
    if (!directoryExists) {
        await this.mkdir(targetDirectory.slice(0, -1)); // Remove the trailing slash, otherwise it would fail
    }
    await this.RNFS.writeFile(target, content, encoding);
} else {
    // Use RN-Fetch-Blob for directly writing
    this.RNFetchBlob.fs.writeFile(...)
}
```

I'm now testing the RNFS RN-Windows implementation after I sorted out the RN-Gesture-Handler + React-Navigation problems.
Should be able to reach the Dashboard tomorrow and test some more RNFS functionalities.

avmoroz:
Hi @creambyemute , I modified the .exists method in the Windows implementation to account for one trailing slash at the end of the path.

I checked the Android implementation, and found that you can have an almost unlimited number of trailing slashes after an existing file and it will be registered. I felt that one trailing forward slash or backslash was reasonable, and more than that was unnecessary. I could not check the iOS implementation because I do not have the necessary devices.

Please let me know if I ought to account for more than one trailing slash.

Thank you for your feedback.

creambyemute:
Hey @avmoroz 
I honestly think one should be enough for the start. If more trailing slashes are passed it's often somewhat of a user/developer error?

I have just tested it and it seems to work fine without the above workaround on `.exists`.

Next question: `.mkdir` also fails without removing the trailing slash, I have looked at the code and just saw the reference to Apples `createDirectoryAtPath` method which seems fine to me. 
But does `createDirectoryAtPath` also fail on a trailing slash? Android is probably going to successfully create the directory?

Thanks a lot for your effort so far!

avmoroz:
Hey @creambyemute ,

I also checked and .mkdir to behaves the same way on Android. I've made a commit to account for this, and hopefully the behavior is consistent across iOS and Android.

I'm looking forward to your feedback on the other features!

creambyemute:
Hey @avmoroz,

I have tested RN-FS + Redux-Persist-FS-Storage with Windows and the trailing slash didn't seem to bother anymore. 
I'll test iOS as well (through redux-persist-fs-storage) soon after resolving the build error I have there since updating RN-Config.

I guess tomorrow or on monday/tuesday I'll come around to replace the  downloading & uploading files method stubs  which would additionally include copyFile, moveFile & unlink.

creambyemute:
Hey @avmoroz 

So I got around to test some more functionalities like downloadFile, copyFile, moveFile and unlink and they seemed to work fine for me.

I've found one more odd behaviour: `.appendFile` on iOS and Android seems to create the file if it doesn't exist, while the CPP implementation doesn't.

I've had to workaround it for windows by first checking&creating the targetDirectory and then also checking&creating the file.

```javascript
async function appendFile(
        filepath: string,
        targetDirectory: string,
        contents: string,
        encodingOrOptions?: string
    ): Promise<void> 
{
        if (Platform.OS === "windows") {
            const doesTargetDirectoryExist = await RNFS.exists(targetDirectory);
            if (!doesTargetDirectoryExist) {
                await RNFS.mkdir(targetDirectory);
            }
            const doesFileExist = await RNFS.exists(filepath);
            if (!doesFileExist) {
                await RNFS.writeFile(filepath, "", encodingOrOptions);
            }
        }
        return await this.RNFS.appendFile(filepath, contents, encodingOrOptions);
}
```

I'm also getting an Error on writeFile at that point, but that's only happening on Login and probably because of multiple/simultaneuos not awaited logging calls in my app. Doesn't happen after the login anymore (because the file is created already at that point).
This error does not appear on iOS/Android.
```
Error: Invalid pointer
    at Object.promiseMethodWrapper [as writeFile] (VM4 index.bundle:2386)
    at Object.writeFile
```

avmoroz:
Hey @creambyemute,

I modified appendFile() to account for the inconsistency between this implementation and the Android and iOS implementations. Please let me know if this error also comes up with this implementation. 

Sorry it took a while to get back to you.

creambyemute:
Hey @avmoroz No worries for the delay. I'll test the change soon, fixing up some other mvp features in the app for windows right now. 

Will let you know after christmas.

Have some good holidays 👍

exotexot:
Hello everyone and happy new year.

Any updates on this. Really want to see this merged :)

stmoy:
Happy new year to you too, @exotexot!

@asklar - do you sign off on this?

@itinance - once asklar signs off, is this good to be merged or are there outstanding things to address?

creambyemute:
Hey @avmoroz and @stmoy 

I have tested a bit more and found something weird when I was trying out the windows implementation of react-native-sketch-canvas:

If I call `.exists(...)` with the `Picture Library` folder (`Drive:\Users\{user}\Pictures\`) or a file path within it (`Drive:\Users\{user}\Pictures\RNSketchCanvas/somePicture.jpg`) RN-FS crashes the app.

read, copy and moveFile do work on the Pictures Library folder though.

jaimecbernardo:
Checking the implementation, it looks like `read`, `copy` and `moveFile` are using `winrt::Windows::Storage` implementations while `.exists` relies on `std::filesystem`.

Is it possible that `std::filesystem` APIs don't work correctly on paths outside the Application's sandbox, such as the Pictures Library and `winrt::Windows::Storage` would have to be used to take into account manifest capabilities checks, if the file was previously picked by a FilePicker, etc...?

stmoy:
I'll let @avmoroz chime in, but my hunch is that @jaimecbernardo is correct - since RNW builds a sandboxed UWP app, I'd imagine that `std::filesystem` wouldn't work - but I'm just guessing.

asklar:
yes, std::filesystem APIs that touch the filesystem will crash the app

avmoroz:
Hey everybody, 

Sorry for the late response.

@creambyemute , I just pushed a commit to address this particular issue. I ran my unit tests and tried it out with my demo app, but please let me know if this implementation of exists() does not resolve your issue or raises new ones for you.

As @jaimecbernardo , @stmoy , and @asklar have pointed out, using std::filesystem to touch files and folders outside the sandbox will cause a crash. I looked through my implementations, and found that appendFile(), mkdir(), stat(), and unlink() have the potential to do that (they utilize std::filesystem::create_directories(), std::filesystem::is_directory(), and std::filesystem::exists). I'll try have new implementations without those std::filesystem functions within the next two days.

avmoroz:
Hey everybody,

I removed all references to std::filesystem::create_directories(), std::filesystem::is_directory(), and std::filesystem::exists() were removed from appendFile(), mkdir(), stat(), and unlink(). I ran some unit tests and manual tests, and it seems like all the functionality is retained.

If there are any issues, concerns, or comments, I'm happy to hear them out and make modifications as necessary

FS.common.js
```diff
@@ -205,6 +206,13 @@ var RNFS = {
     return RNFSManager.copyFile(normalizeFilePath(filepath), normalizeFilePath(destPath), options).then(() => void 0);
   },
 
+  // Windows workaround for slow copying of large folders of files
+  copyFolder(filepath: string, destPath: string): Promise<void> {
+    if(isWindows) {
+      return RNFSManager.copyFolder(normalizeFilePath(filepath), normalizeFilePath(destPath));
+    }
```
asklar, line 213: what happens on non-Windows? or does this need to somehow be marked as Windows-only ?

windows/RNFS.Tests/ReactNativeCxxTests/JsonJSValueReader.h
asklar: @avmoroz just making sure you saw this comment, do we need this custom json parser?

windows/RNFS/RNFSManager.cpp
```diff
@@ -0,0 +1,1115 @@
...
+        StorageFolder folder{ co_await StorageFolder::GetFolderFromPathAsync(path.parent_path().wstring()) };
+        co_await folder.CreateFolderAsync(path.filename().wstring(), CreationCollisionOption::FailIfExists);
+
+        promise.Resolve();
+    }
+}
+catch (const hresult_error& ex)
+{
+    // "Unexpected error while making directory."
+    promise.Reject( winrt::to_string(ex.message()).c_str() );
```
asklar, line 208: nit can probably avoid the `c_str()`

windows/RNFS/RNFSManager.cpp
```diff
@@ -0,0 +1,1115 @@
...
+        auto target{ co_await folder.GetItemAsync(path.filename().wstring()) };
+        co_await target.DeleteAsync();
+
+        promise.Resolve();
+    }
+}
+catch (const hresult_error& ex)
+{
+    hresult result{ ex.code() };
+    if (result == 0x80070002) // FileNotFoundException
```
asklar, line 368: use HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND)

asklar: same in other places

windows/RNFS/RNFSManager.cpp
```diff
@@ -0,0 +1,1115 @@
...
+    promise.Resolve(winrt::to_string(base64Content));
+}
+catch (const hresult_error& ex)
+{
+    hresult result{ ex.code() };
+    if (result == 0x80070002) // FileNotFoundException
+    {
+        promise.Reject(RN::ReactError{ "ENOENT", "ENOENT: no such file or directory, open " + filepath });
+    }
+    else if (result == 0x80070005) // UnauthorizedAccessException
```
asklar, line 444: `E_ACCESSDENIED`

windows/RNFS/RNFSManager.cpp
```diff
@@ -0,0 +1,1115 @@
...
+        fileInfo["ctime"] = winrt::clock::to_time_t(item.DateCreated());
+        fileInfo["mtime"] = winrt::clock::to_time_t(properties.DateModified());
+        fileInfo["size"] = std::to_string(properties.Size());
+        fileInfo["type"] = item.IsOfType(StorageItemTypes::Folder) ? 1 : 0;
+        promise.Resolve(fileInfo);
+    }
+}
+catch (...)
+{
+    promise.Reject(RN::ReactError{ "ENOENT", "ENOENT: no such file or directory, open " + filepath });
```
asklar, line 484: not necessarily this error though, right?

windows/RNFS/RNFSManager.cpp
```diff
@@ -0,0 +1,1115 @@
...
+    Streams::IBuffer buffer{ co_await FileIO::ReadBufferAsync(file) };
+
+    auto hashedBuffer{ provider.HashData(buffer) };
+    auto result{ winrt::to_string(Cryptography::CryptographicBuffer::EncodeToHexString(hashedBuffer)) };
+
+    promise.Resolve(result);
+}
+catch (const hresult_error& ex)
+{
+    hresult result{ ex.code() };
```
asklar, line 593: extract this out to its own helper function and calling it from the different catch clauses so you don't repeat the same code (and ensures the string messages are consistent as well as the set of errors handled)

windows/RNFS/RNFSManager.cpp
```diff
@@ -0,0 +1,1115 @@
...
+
+
+void RNFSManager::touch(std::string filepath, int64_t mtime, int64_t ctime, bool modifyCreationTime, RN::ReactPromise<std::string> promise) noexcept
+try
+{
+
+    std::filesystem::path path(filepath);
+    path.make_preferred();
+    auto s_path{ path.c_str() };
+    LPCWSTR actual_path{ s_path };
```
asklar, line 836: is this needed? if so drop the L, just PCWSTR is fine (L is a now deprecated / no-op type prefix from 16 bit Windows :)  )

windows/RNFS/RNFSManager.cpp
```diff
@@ -0,0 +1,1115 @@
...
+                    { "contentLength", contentLength.Type() == PropertyType::UInt64 ? RN::JSValue(contentLength.Value()) : RN::JSValue{nullptr} },
+                    { "headers", std::move(headersMap) },
+                });
+        }
+
+        uint64_t totalRead{ 0 };
+
+        std::filesystem::path fsFilePath{ filePath };
+
+		StorageFolder storageFolder{ co_await StorageFolder::GetFolderFromPathAsync(fsFilePath.parent_path().wstring()) };
```
asklar, line 933: nit, fix indentation

windows/RNFS/RNFSManager.h
```diff
@@ -0,0 +1,180 @@
...
+{
+    CancellationDisposable() = default;
+	CancellationDisposable(winrt::Windows::Foundation::IAsyncInfo const& async, std::function<void()>&& onCancel) noexcept;
+
+	CancellationDisposable(CancellationDisposable&& other) noexcept;
+	CancellationDisposable& operator=(CancellationDisposable&& other) noexcept;
+
+	CancellationDisposable(CancellationDisposable const&) = delete;
+	CancellationDisposable& operator=(CancellationDisposable const&) = delete;
+
```
asklar, line 26: nit fix indentation

windows/RNFS/RNFSManager.h
```diff
@@ -0,0 +1,180 @@
...
+    winrt::Windows::Foundation::IAsyncAction ProcessUploadRequestAsync(RN::ReactPromise<RN::JSValueObject> promise, RN::JSValueObject& options,
+        winrt::Windows::Web::Http::HttpMethod httpMethod, RN::JSValueArray const& files, int32_t jobId, uint64_t totalUploadSize);
+
+    winrt::fire_and_forget copyFolderHelper(
+        winrt::Windows::Storage::StorageFolder src,
+        winrt::Windows::Storage::StorageFolder dest) noexcept;
+
+    constexpr static int64_t UNIX_EPOCH_IN_WINRT_INTERVAL = 11644473600 * 10000000;
+     
+    const std::map<std::string, std::function<CryptographyCore::HashAlgorithmProvider()>> availableHashes{
```
asklar, line 169: this can be constexpr / static const? also no need for a map where an unordered_map or even a `vector<pair<string, function>>` will do :) it's only 5 elements

asklar (approved):

FS.common.js
avmoroz: This particular function needs to be marked as Windows-only. It exists as a workaround to make sure that copying directories with contents is done quickly. This can be done via JavaScript, but I suspect that the time it takes to make all the necessary calls increases the time needed to transfer all those files and contents from a fraction of a second to 30-40 seconds easily in a 60 megabyte folder with 60 different subfolders and files.

Is there a good (or at least, acceptable) way to do that?

asklar: I'm not sure if there is a pattern for this but I want to make sure this doesn't break apps running on non-Windows devices

windows/RNFS.Tests/ReactNativeCxxTests/JsonJSValueReader.h
avmoroz: To be completely honest, I asked @vmoroz about a way to create unit tests for my new implementations of RNFS functions, and he recommended I test the functions like this. This custom json parser is solely there for the unit tests.

windows/RNFS/RNFSManager.cpp
avmoroz: Unfortunately, promise.Reject() does not accept a std::string. It fails to build when I do not use the c_str(). Would it be better to return ReactErrors instead?

windows/RNFS/RNFSManager.cpp
avmoroz: Replaced all occurances of that in my most recent commit.

FS.common.js
avmoroz: Because it's only intended for Windows, would returning a promise rejection stating that this function is not available on other platforms be acceptable for other platforms?

windows/RNFS/RNFSManager.cpp
avmoroz: Just pushed a commit that implements that change.

windows/RNFS/RNFSManager.cpp
avmoroz: Tested this method using just PCWSTR, and I'm still getting the same results. I made that change in a recent commit.

On the topic of whether or not it is needed...
Android doesn't allow for users to update the file creation time, but iOS does. I thought it would be better to try and replicate the same functionality of iOS for this method.

I couldn't find a way to update the file creation time in WinRT, so I decided to use CreateFile2() from Win32. 

This Win32 method should still work across all Windows 10 devices. Should I remove the ability to update the creation time of a file so that only WinRT is used?

windows/RNFS/RNFSManager.h
avmoroz: Recent commit fixes this.

windows/RNFS/RNFSManager.cpp
avmoroz: Fixed in recent commit.

creambyemute:
@avmoroz I've got around to test the new implementations. Checking if the signature from rn-sketch-canvas exists works now. Copying/moving worked as well.

Unfortunately, directory creation seems to be working quite different (regression?) now than before:
* I have to call mkdir before doing any copy/move actions, otherwise it complains
* If I call mkdir on an already existing directory, it complains, so you have to add a !exist check first.

I think Android handles this gracefully. Should one of those behaviours be expected or do we have to adapt our JS implementations to check for existence of directories before mkdir/copyfile/movefile etc?

avmoroz:
Hey @creambyemute ,

I did some investigation, and your issue with directory creation was an error on my part. mkdir() is not supposed to provide a promise rejection if the directory in question already exists or was not able to be created. It no longer provides a rejection as of this most recent commit.

I also dug around and realized that my implementation of mkdir() did not create parent directories if of the new directory if they did not initially exist. Hopefully that hasn't been an issue for you, but that has also been fixed in the most recent commit.

Just to clarify: was your issue with mkdir() specifically, or were you having issues with copyfile() and movefile() as well?

creambyemute:
@avmoroz I noticed it while trying to moveFile/copyFile to a not yet existing directory. I then added  mkdir calls and noticed that it then complains about the directory existing already.

exotexot:
Can we talk about what the next steps might be in this regard?

stmoy:
Ping @avmoroz - looks like there is a merge conflict now. Can you please resolve and identify if there is any outstanding work? If there isn't any more work, is this good to get merged?

creambyemute:
I've tried upgrading our Project to 0.64 and it fails building on `react-native-fs`. I then tried it with a fresh 0.64 init project and added rn-fs in there and it fails building as well on react-native-fs.

I create an issue in the rn-w repository first, but I think it rather belongs here after some more investigation: https://github.com/microsoft/react-native-windows/issues/7585

The `npx react-native run-windows` output I get is:

```
5>Done Building Project "C:\Users\thoma\dev\react\RN064Init2\node_modules\@terrylinla\react-native-sketch-canvas\windows\RNSketchCanvas\RNSketchCanvas.vcxproj" (default targets).
    11>C:\Users\thoma\dev\react\RN064Init2\node_modules\react-native-windows\Microsoft.ReactNative.Cxx\JSI\JsiApiContext.cpp(9,40): error C4430: missing type specifier - int assumed. Note: C++ does not support default-int [C:\Users\thoma\dev\react\RN064Init2\node_modules\react-native-fs\windows\RNFS\RNFS.vcxproj]
    11>C:\Users\thoma\dev\react\RN064Init2\node_modules\react-native-windows\Microsoft.ReactNative.Cxx\JSI\JsiApiContext.cpp(9,29): error C2146: syntax error: missing ';' before identifier '__ImageBase' [C:\Users\thoma\dev\react\RN064Init2\node_modules\react-native-fs\windows\RNFS\RNFS.vcxproj]
    11>C:\Users\thoma\dev\react\RN064Init2\node_modules\react-native-windows\Microsoft.ReactNative.Cxx\JSI\JsiApiContext.cpp(34,5): error C2065: 'HMODULE': undeclared identifier [C:\Users\thoma\dev\react\RN064Init2\node_modules\react-native-fs\windows\RNFS\RNFS.vcxproj]
    11>C:\Users\thoma\dev\react\RN064Init2\node_modules\react-native-windows\Microsoft.ReactNative.Cxx\JSI\JsiApiContext.cpp(34,13): error C2146: syntax error: missing ';' before identifier 'currentDllHanlde' [C:\Users\thoma\dev\react\RN064Init2\node_modules\react-native-fs\windows\RNFS\RNFS.vcxproj]
    11>C:\Users\thoma\dev\react\RN064Init2\node_modules\react-native-windows\Microsoft.ReactNative.Cxx\JSI\JsiApiContext.cpp(34,13): error C2065: 'currentDllHanlde': undeclared identifier [C:\Users\thoma\dev\react\RN064Init2\node_modules\react-native-fs\windows\RNFS\RNFS.vcxproj]
    11>C:\Users\thoma\dev\react\RN064Init2\node_modules\react-native-windows\Microsoft.ReactNative.Cxx\JSI\JsiApiContext.cpp(34,49): error C2061: syntax error: identifier 'HMODULE' [C:\Users\thoma\dev\react\RN064Init2\node_modules\react-native-fs\windows\RNFS\RNFS.vcxproj]
    11>C:\Users\thoma\dev\react\RN064Init2\node_modules\react-native-windows\Microsoft.ReactNative.Cxx\JSI\JsiApiContext.cpp(35,101): error C2065: 'currentDllHanlde': undeclared identifier [C:\Users\thoma\dev\react\RN064Init2\node_modules\react-native-fs\windows\RNFS\RNFS.vcxproj]
    11>C:\Program Files (x86)\Windows Kits\10\bin\10.0.18362.0\XamlCompiler\Microsoft.Windows.UI.Xaml.Common.targets(482,5): error MSB4181: The "CompileXaml" task returned false but did not log an error. [C:\Users\thoma\dev\react\RN064Init2\node_modules\react-native-fs\windows\RNFS\RNFS.vcxproj]
    11>Done Building Project "C:\Users\thoma\dev\react\RN064Init2\node_modules\react-native-fs\windows\RNFS\RNFS.vcxproj" (default targets) -- FAILED.
     2>Done Building Project "C:\Users\thoma\dev\react\RN064Init2\windows\rn064init2\rn064init2.vcxproj" (default targets) -- FAILED.
    12>Done Building Project "C:\Users\thoma\dev\react\RN064Init2\windows\rn064init2\rn064init2.vcxproj.metaproj" (default targets) -- FAILED.
     1>Done Building Project "C:\Users\thoma\dev\react\RN064Init2\windows\rn064init2.sln" (default targets) -- FAILED.

```

windows/RNFS/pch.h
```diff
@@ -0,0 +1,3 @@
+﻿#pragma once
+#include <winrt/Windows.Foundation.h>
+#include <winrt/Windows.Foundation.Collections.h>
```
asklar, line 3: please add 
```cpp
#include <windows.h>
```

Since RNFS uses some jsi headers that rely on windows.h (see the issue that @creambyemute ran into)

avmoroz: Just updated this.

avmoroz:
Sorry for the delay, everybody, I had to take care of some external issues.

It seems like the merge conflict was due to a version upgrade. I resolved that recently. I updated the example project to utilize React Native version 0.64.0 and got it seems to be working fine.  

If any issues persist or come up, feel free to let me know. I'd love to get this PR merged, but I also appreciate everyone who's pointed out any shortcomings they run into!

windows/RNFS/pch.h
asklar: thanks a lot @avmoroz!

asklar:
@itinance any concerns merging this?

vincentjames501:
Woot! Thanks @avmoroz . Would love to see this one get merged in. Super useful!

tero-paananen:
I am using this in our app and i had to add quite many null checks `contentLength != nullptr && contentLength.Type()` into `RNFSManager::ProcessDownloadRequestAsync` when module is emiting events. Content length value seems to be null at least when response is 404

windows/RNFS/RNFSManager.cpp
```diff
@@ -0,0 +1,1138 @@
...
+        winrt::hstring directoryPath, fileName;
+        splitPath(filepath, directoryPath, fileName);
+        StorageFolder folder{ co_await StorageFolder::GetFolderFromPathAsync(directoryPath) };
+        if (fileName.size() > 0) {
+            co_await folder.GetItemAsync(fileName);
+        }
+        promise.Resolve(true);
+    }
+}
+catch (const hresult_error& ex)
```
jeanmaried, line 425: This catch is never called if the filepath points to a non-existent path. These are the type of exceptions in the output:
```
Exception thrown at 0x00007FFED89D4B89 (KernelBase.dll) in MyApp.exe: WinRT originate error - 0x80070002 : 'The system cannot find the file specified.'.
Exception thrown at 0x00007FFED89D4B89 in MyApp.exe: Microsoft C++ exception: winrt::hresult_error at memory location 0x0000003BE39FD5F0.
```

windows/RNFS/RNFSManager.cpp
```diff
@@ -0,0 +1,1138 @@
...
+    splitPath(filepath, directoryPath, fileName);
+
+    StorageFolder folder{ co_await StorageFolder::GetFolderFromPathAsync(directoryPath) };
+    StorageFile file{ co_await folder.GetFileAsync(fileName) };
+
+    Streams::IBuffer buffer{ co_await FileIO::ReadBufferAsync(file) };
+    winrt::hstring base64Content{ Cryptography::CryptographicBuffer::EncodeToBase64String(buffer) };
+    promise.Resolve(winrt::to_string(base64Content));
+}
+catch (const hresult_error& ex)
```
jeanmaried, line 461: This catch is also not reached if the file path points to non-existent file. These are the type of exceptions in the output:
```
Exception thrown at 0x00007FFED89D4B89 (KernelBase.dll) in MyApp.exe: WinRT originate error - 0x80070002 : 'The system cannot find the file specified.'.
Exception thrown at 0x00007FFED89D4B89 in MyApp.exe: Microsoft C++ exception: winrt::hresult_error at memory location 0x0000003BE39FD5F0.
```

jeanmaried (commented):
Testing out the `exists` and `readFile` methods I found that they fail and crash the app when the file path points to a non-existent file. When the error is thrown, the catch that is supposed to return a `promise.Resolve(false)` is never called.

windows/RNFS/RNFSManager.cpp
tero-paananen: I noticed yesterday also these `The system cannot find the file specified` errors in my tests.

tero-paananen:
Is there error in error handling for example in `RNFSManager::exists` ?

```
catch (const hresult_error& ex)
{
    hresult result{ ex.code() };
    if (result == HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND)) {
        promise.Resolve(false);
    }
    // "Failed to check if file or directory exists.
    promise.Reject(winrt::to_string(ex.message()).c_str());
}
```

code calls promise twice, first `promise.Resolve(false);` and then `promise.Reject(winrt::to_string(ex.message()).c_str());` if error is that `ERROR_FILE_NOT_FOUND`

windows/RNFS/RNFSManager.cpp
tero-paananen: But when i debug code it seems that it goes to catch and handles it. There is still that error in console `Exception thrown at 0x00007FFED89D4B89 (KernelBase.dll) in MyApp.exe: WinRT originate error - 0x80070002 : 'The system cannot find the file specified.'` maybe it is ok?

tero-paananen:
When RNFS receives lot of `Exception thrown at 0x00007FFED89D4B89 (KernelBase.dll) in MyApp.exe: WinRT originate error - 0x80070002 : 'The system cannot find the file specified.`  errors is my Visual Studio debug always stopped. I had to press `Continue` to let app run again. This is not very developer friendly? I had to remove this `winrt:hresult_error` error from list of `Break When Thrown` to not always stop this. I do not know what is right solution for this.
error

exotexot:
This is also a observation I have made. If this could be fixed appropriately, that would be awesome! @avmoroz

emanuelmartin:
Hi,

I have a doubt using this package,

1. I added this module to my project by downloading the dev brach of @avmoroz and adding it to node modules
2. Added `#include "winrt/RNFSManager.h"` into `pch.h`
3. Added `PackageProviders().Append(wintry::RNFSManager::ReactPackageProvider());` before `InitializeComponent();` in `App.cpp`

exotexot:
Any updated happening here? :(

asklar:
@Chiaramooney could you take on driving to get this merged?

raaannzzz:
+1

emanuelmartin:
> @chiaramooney could you take on driving to get this merged?

So happy to hear that 💯

namrog84:
Any potential update on this getting in?  Would love to see it happen!

emanuelmartin:
We need a GIF saying "Merge it mate"

exotexot:
Anyone tested this with a recent react-native version under Windows 11 ?

datacana:
sadly no. example cannot build. does throw errors in react native JsiApiContext.cpp 

Screenshot 2021-11-30 231825

@asklar @avmoroz  any ideas?

asklar:
> sadly no. example cannot build. does throw errors in react native JsiApiContext.cpp
> 
> Screenshot 2021-11-30 231825
> 
> @asklar Alexander Sklar FTE @avmoroz any ideas?

See https://github.com/microsoft/react-native-windows/issues/7585#issuecomment-821626197

windows/RNFS/RNFS.def
```diff
@@ -0,0 +1,3 @@
+﻿EXPORTS
+DllCanUnloadNow = WINRT_CanUnloadNow                    PRIVATE
```
namrog84, line 2: I was testing this PR in our code via a patch and was getting weird error around this file when it wasn't using crlf (it appears to be currently lf)  Visual Studio 2019 compiler errored out on it.

Consider updating to crlf?

namrog84:
@asklar, @avmoroz, @tero-paananen

Is there anything still blocking this branch?  

It'd be awesome to just get it in, and then would be easier to fix bugs/make PRs to it then.

asklar:
> @asklar, @avmoroz, @tero-paananen
> 
> Is there anything still blocking this branch from going in?
> 
> It'd be awesome to just get it in, and then would be easier to fix bugs/make PRs to it then.

No concerns from me to merge and fix bugs separately as the pr is already large. However I don't know how much appetite @itinance might have to merge this (as it's been open for a long time), and if further changes are required to fix bugs, I'd be concerned about the velocity with which changes are accepted. So perhaps a fork might be required.

creambyemute:
I'm currently facing a problem with user-directory paths that contain Umlaute/accentuated characters

Maybe someone has an idea how to handle the std::string --> std::path conversion correctly?

https://github.com/microsoft/react-native-windows/issues/9315

creambyemute:
@asklar @namrog84 @avmoroz FYI
I was testing my std::string --> path  withUmlaute problem today and got the app working with a few rather small changes in RNFS:

I created a convertPath method and adapted the widely used splitPath method as well. 
With a few more changes (using convertPath(...) where `splitPath(...)` isn't used) + a change to `downloadFile(...)` (`path.string()` instead of `path.c_str()`) it started working.


I don't like this solution yet (as it probably isn't correct?) and it can most likely be further narrowed down and simplified.
```cpp
std::filesystem::path RNFSManager::convertPath(const std::filesystem::path originalPath) 
{
    winrt::hstring fPathHString = winrt::to_hstring(originalPath.string());
    std::filesystem::path path{fPathHString.c_str()};
    path.make_preferred();
    return path;
}


void RNFSManager::splitPath(const std::string& fullPath, winrt::hstring& directoryPath, winrt::hstring& fileName) noexcept
{
    std::filesystem::path path(fullPath);
    path.make_preferred();
    std::filesystem::path convertedPath = convertPath(path);

    directoryPath = convertedPath.has_parent_path() ? winrt::to_hstring(convertedPath.parent_path().c_str()) : L"";
    fileName = convertedPath.has_filename() ? winrt::to_hstring(convertedPath.filename().c_str()) : L"";
}

winrt::fire_and_forget RNFSManager::downloadFile(...) noexcept
{
    auto filePath{ winrt::to_hstring(path.string()) };
}

// In mkdir and some others
std::filesystem::path firstPath(pathString);
firstPath.make_preferred();
std::filesystem::path path = convertPath(firstPath);
```

creambyemute:
@namrog84 FYI: I have created a branch (fixUmlautePaths) and PR on https://github.com/wwimmo/react-native-fs if your team wants to try out the changes

taduyde:
when we merge this branch?

shanelile:
Are there any updates to this? Is it blocked by anything?

Issue #843, opened 2020-02-04
RNFS uploadFiles fails at 0% ,uploadProgress will run infinitely and freezes app


reetaramadurg:
var RNFS = require('react-native-fs');
        var uploadUrl = 'myurl'     
        var file = [
            {
                name: 'files',
                filename: fileNames,
                filepath: fileNames,
            }
        ];
        var uploadBegin = (response) => {
            var jobId = response.jobId;
            console.log('UPLOAD HAS BEGUN! JobId: ' + jobId);
        };

        var uploadProgress = (response) => {
            var percentage = Math.floor((response.totalBytesSent / response.totalBytesExpectedToSend)       * 100);
            console.log('UPLOAD IS ' + percentage + '% DONE!');
            };
        console.log(file)
        RNFS.uploadFiles({
            toUrl: uploadUrl,
            files: file,
            method: 'POST',
            headers: {
                'Accept': 'application/json',
            },
            begin: uploadBegin,
            progress: uploadProgress
            }).promise.then((response) => {
            if (response.statusCode == 200) {
                console.log('FILES UPLOADED!'); // response.statusCode, response.headers, response.body
                var res = JSON.parse(response.body)
                console.log(res)
                var checkFileExists = RNFS.DocumentDirectoryPath + '/syncedFiles'
                checkFileExists = checkFileExists + "/" + res.FileName
                RNFS.exists(checkFileExists)
                    .then((doesFileExists) => {
                        console.log("doesFileExists " + doesFileExists)
                        if (doesFileExists == true) {
                            RNFS.unlink(checkFileExists)
                                .then(() => {
                                    console.log("fileDeleted " + checkFileExists)
                                    this.moveFile(fileNames, res.FileName)
                                })
                        } else {
                            this.moveFile(fileNames, res.FileName)
                        }
                    })

            } else {
                console.log('SERVER ERROR');

            }
        })
            .catch((err) => {
                if (err.description === "cancelled") {
                    // cancelled by user
                }
                console.log(err);
            });
    }

hank121314:
Which platform did you use? Android or IOS?

Donkey1998:
Android

PR #954, opened 2020-11-18
Switch to using channels for byte copying.

ball-hayden:
This avoids having to deal with buffer allocation ourselves, and should be more efficient.

I believe this resolves https://github.com/itinance/react-native-fs/issues/843


itinance:
Many thanks man!

PR #963, opened 2020-11-27
Add basic support for Android content URIs in stat

stevenremot:
Add a basic implementation of stat for content URIs that returns what
it can return: the file length.

This also raises the Android SDK API requirements of this method from 1 to 3, which is... fine I guess ?

If you see anything else I can return for a content URI I will be glad to add it.

### DIFF
--- a/android/src/main/java/com/rnfs/RNFSManager.java
+++ b/android/src/main/java/com/rnfs/RNFSManager.java
@@ -5,6 +5,7 @@ import android.content.res.AssetManager;
 import android.database.Cursor;
 import android.net.Uri;
 import android.os.AsyncTask;
+import android.os.Build;
 import android.os.Environment;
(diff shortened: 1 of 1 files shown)


viveking (changes requested):
It seems you not providing.
- ctime , mtime, type[MIME], orginalFilepath 

Please handle the same.

stevenremot:
Thanks for the feedback! It doesn't look as straightforward as the non-content schemes, but I will have a look at it.

stevenremot:
It's been a month and I haven't been able to take a look at it. I'm not sure I will be able to in the near future. So feel free to close this PR if you want to keep the project clean.

itinance:
Alright. I'm going to close this as you suggested. Pls feel free everybody to bring this to life later.

PR #986, opened 2021-02-28
macOS support

Bunkerbewohner:
Hi! This allows using `react-native-fs` on macOS.

I confirmed that it works with a react-native app for macOS based on https://microsoft.github.io/react-native-windows/.
Though I haven't used all of the API methods yet, just the basics like `readdir`, `readFile` and so on.

I didn't get the integration tests to build, but that is I think unrelated to my changes.


itinance:
Brilliant! Many thanks!

Bunkerbewohner:
Wow, quick reaction! Thank you for this great library! 👏

Issue #997, opened 2021-04-13
Could not resolve dependency: peer react-native@"^0.59.5" from react-native-fs@2.17.0


keremoge:
Because the react-native version specified in [package.json](https://github.com/itinance/react-native-fs/blob/master/package.json), when you try `npm i` with npm version **7.*** you get following error because not using specified react-native version. There is a workaround which is using `npm i --legacy-peer-deps`.

```
Could not resolve dependency:
npm ERR! peer react-native@"^0.59.5" from react-native-fs@2.17.0
```

package.json:
```
  "peerDependencies": {
    "react-native": "^0.59.5",
    "react-native-windows": "^0.57.2"
  }
```

keremoge:
I think `peerDependencies`  should be removed from [package.json](https://github.com/itinance/react-native-fs/blob/master/package.json)

hussainarthuna:
Facing the same problem

danielkoh99:
Same issue

itinance:
> I think `peerDependencies` should be removed from [package.json](https://github.com/itinance/react-native-fs/blob/master/package.json)

do you want to submit a PR? Thx :) 
(Or someone else?)

keremoge:
I submitted a PR #1001  👍

nkulic:
Same issue here, 

`npm ERR! Found: react-native@0.63.4
npm ERR! node_modules/react-native
npm ERR!   react-native@"^0.63.4" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer react-native@"^0.59.5" from react-native-fs@2.17.0
npm ERR! node_modules/react-native-fs
npm ERR!   react-native-fs@"^2.17.0" from the root project
npm ERR! 
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.`

mans17:
The workaround submitted does not seem to work here.
I tested : 

`npm install react-native-fs  --legacy-peer-deps` 

And I got : 
`npm ERR! subset is not a function`

/home/user/.npm/_logs/2021-05-02T09_01_39_772Z-debug.log
```
53 verbose stack TypeError: subset is not a function
53 verbose stack     at Arborist.[saveIdealTree] (/usr/lib/node_modules/npm/node_modules/@npmcli/arborist/lib/arborist/reify.js:971:37)
[...]

PR #1001, opened 2021-04-25
remove peerDependencies

keremoge:
remove `peerDependencies` from [package.json](https://github.com/itinance/react-native-fs/blob/master/package.json) because of `npm 7+ `
#997


Overtorment:
@itinance are you going to merge it? want to use `react-native-fs` ith npm@7

PR #1004, opened 2021-05-03
Update Android's UploadFiles boundary to use random UUID 

nikoWah:
Changes made in UploadFiles for android:
- WebGateway security setting blocked file uploads from Android as it found the fixed asterisks boundary string to be duplicated requests.
- Update boundary to use random UUID instead of using fixed asterisks to be consistent with iOS implementation.


bachbonglong:
Pls merge this code. I need boundary not *****

PR #1029, opened 2021-07-26
Made read() use 64-bit position rather than 32 bit for large files

### LINKED ISSUE #1027: "Read" problem for overly large files
I have an app performing uploads of large files, using "read" to extract Base64 chunks and upload them; today found an issue with my uploads freezing beyond a certain point, and it's due to the "position" being larger than int_max, so when cast to (int) seek file to offset fails. 
Pretty simple fix (on iOS at least) - instead of casting the variables to integers, they should be cast to longs, which works for NSIntegers.  Still working on Android, but will post a pull request (at least for "read") when that's working

fwestling:
Resolves #1027

### DIFF
--- a/RNFSManager.m
+++ b/RNFSManager.m
@@ -323,7 +323,7 @@ RCT_EXPORT_METHOD(read:(NSString *)filepath
     }
 
     // Seek to the position if there is one.
-    [file seekToFileOffset: (int)position];
+    [file seekToFileOffset: (long)position];
 
     NSData *content;
     if ((int)length > 0) {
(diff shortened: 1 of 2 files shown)


elliottkember:
I just hit this same problem reading a really big file in React Native. This patch looks good - is there any reason not to merge this?

RNFSManager.m
```diff
@@ -323,7 +323,7 @@ + (BOOL)requiresMainQueueSetup
     }
 
     // Seek to the position if there is one.
-    [file seekToFileOffset: (int)position];
```
thomasttvo, line 326: It seems like we can just remove the `*` from line 296 and 297 then remove the casting here and line 330 and we'll be ok

Issue #933, opened 2020-09-22
Chinese garbled characters when uploading files


qcmoke:
**How to solve the problem of garbled Chinese characters when uploading files？**

Project dependencies：
>    "react": "16.13.1",
>    "react-native": "0.63.2",
>    "react-native-fs": "^2.16.6",

PR #1045, opened 2021-10-11
[TangChange] #933 fix: chinese garbled characters

daltachange:


itinance:
Thanks for contributing. PLs take a look at #1088, where UTF-8 encoding was addressed. Both PR together  will cause merge conflicts and problably side effects.

PR #1086, opened 2022-03-24
Fix progressCallback not get called when call downloadFile on iOS platform

zhima:
On iOS platform, when I call `downloadFile` method without setting the `begin` callback param, the `progress` callback not get called anyway.

In **Downloader.m**  file, `progressCallback` get called only when `statusCode` equal to 200, and `statusCode` is assigned value only when there is `beginCallback`.

So I put the `statusCode`'s assigning before `beginCallback`'s determination,  so that `statusCode` will be assigned value without  `beginCallback`.


siisee11:
Thanks.

PR #1088, opened 2022-04-02
Support unicode encode request when uploading on Android

Silvercast:
`uploadFiles` with unicode `fields?: Fields` is broken on Android but works on iOS. This PR ensures the fields are properly encoded on Android, otherwise we end up with weird ? characters.


Silvercast:
> Especially in conjunction with this PR: https://github.com/itinance/react-native-fs/pull/1045/files #1045

@itinance  #1045 should work. I don't know if `getBytes()` support all locales. I can test once #1045 get merged.

Issue #760, opened 2019-09-18
progress not work when downloadFile


xclidongbo:
version:     "react-native-fs": "2.13.3"
platform: iOS

code:

```
let documentPath = RNFS.DocumentDirectoryPath; 
    let zipPath = documentPath+'/app.zip'
    let options = {
      fromUrl: downLoadUrl,
      toFile: zipPath,
      background: true,
      progressDivider: 5,
      discretionary: true,
      cacheable: true,
      begin: (res) => {
        console.log('begin', res);
      },
      progress: (res) => {
        console.log('progress', res)
      }
    }
    let ret = RNFS.downloadFile(options);
    ret.promise.then(res=>{
      console.log('====================================');
      console.log('complete',res);
      console.log('====================================');
    }).catch(err=>{
      console.log(err);
    })

```

xclidongbo:
试了两个版本,iOS并不支持下载进度. 这个版本很长时间不维护, 估计凉了.

xclidongbo:
I think it is necessary to fix this problem.

sebinq:
Something new here?

sebinq:
@xclidongbo it looks like when i provide `begin` callback together with `progress` (even when begin is noop), then progress is working correctly.

oisu:
@sebinq you saved my day, thanks!

LuisRodriguezLD:
mine too! thanks @sebinq but funny, I remember this working without begin

sebinq:
@LuisRodriguezLD yep, it was working before just fine. I don't know what what has changed :D But it is good that there is this hack which make it works again.

Charlicus:
@sebqq Thanks !, Saved my evening :-p

sebqq:
Glad to help 🍻

amerllica:
Thanks @sebqq for your awesome help.

I don't know why still it isn't fixed, by using a default `noop` function this issue would be fixed.

raahimkhan:
Facing the same issue right now with `progress` not working with uploading files. Can anyone please help. I have added the `begin` callback as well:

```javascript
const fields = {
            "x-goog-date": credentials.x_goog_date,
            "success_action_status": credentials.success_action_status,
            "x-goog-algorithm": credentials.x_goog_algorithm,
            "content-encoding": credentials.content_encoding,
            "x-goog-credential": credentials.x_goog_credential,
            "acl": credentials.acl,
            "key": credentials.key,
[...]

PR #1091, opened 2022-04-08
Update README.md

AlexPlekhov:
Add a comment for begin prop for downloadFile method


AlexPlekhov:
Based on https://github.com/itinance/react-native-fs/issues/760

PR #1116, opened 2022-07-10
how to install the fixUmlautePaths branch with yarn add?

gangli66:
I want to use the fixUmlautePaths branch in my project because I need to link it to react-native-windows which could not autolink. But how need I write my command with "yarn add ??????"?
I wrote it just like yarn add https://github.com/wwimmo/react-native-fs/tree/fixUmlautePaths
and https://github.com/wwimmo/react-native-fs#fixUmlautePaths， but there was an error like SSL_open 10054 or invalid tar header.


gangli66:
Thanks for your answer. However, how can I take write access? I really need it to finish my program with react-native-windows

PR #1164, opened 2023-02-07
Fix types for StatResult's ctime and mtime

jlpiedrahita:
According to the [docs](https://github.com/itinance/react-native-fs#statfilepath-string-promisestatresult), type for `ctime` and `mtime` in `StatResult` should be `Date` not `number`.


tiagomsmagalhaes:
@itinance can you expedite this simple PR?

vincon09:
ctime and mtime are giving the same times, why so? please help

PR #1244, opened 2024-11-17
Update build.gradle

enisinanaj:
Updated build.gradle including the namespace


arcaela:
> Updated build.gradle including the namespace

This repo have not maintance.

PR #1259, opened 2025-12-23
fix(android): prevent crash when Promise.reject receives null code

nitinkumarnk-me:
### Fix Android crash when Promise.reject receives null code

RNFS may call `Promise.reject()` with a null `code` parameter, which causes a
`NullPointerException` in `PromiseImpl.reject` on Android.

This patch ensures a non-null error code is always provided when rejecting
promises, preventing native crashes without changing JS behavior.


OmarMohammedFoad (approved):
i see that this is the  solution in my case

Kiryl-Sadko (approved):

kOOnzTe (approved):

